PR #16, opened 2026-01-08
chore(ci): fix ci serurity and add build wheels for test pypi

Cuiyus:


.github/workflows/build_test_wheel.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: (Optional) Install and test from TestPyPI
+        if: success() && github.event_name == 'workflow_dispatch'
+        run: |
+          # Create a clean venv
+          python -m venv test_env
+          source test_env/bin/activate
+          pip install --upgrade pip
+          # Install from TestPyPI (must allow pre-releases if version has dev/alpha)
+          pip install --index-url https://test.pypi.org/simple/ \
+                      --extra-index-url https://pypi.org/simple/ \
```
JalinWang: Why do we need to set "--extra-index-url" here?

PR #26, opened 2026-01-16
fix: fix and refact segment fetch perf

chinaux:
When a segment is opened in use_map=true, read_only=true mode, the fetch_perf function will take effect. If the parameter columns contain a LOCL_ROW-ID that is not the last one, it will cause an exception.


egolearner (approved):
LGTM

PR #83, opened 2026-02-09
[enhance] buffer storage

### LINKED ISSUE #64: [Enhance]: Buffer Storage
### Affected Component

buffer manager, indexes who use buffer storage.

### Current Behavior

Current buffer storage performance is poor, limited by excessive concurrency lock overhead.
Lack of write capability.

### Desired Improvement

Optimize the LRU implementation using a data structure with strong concurrency capabilities to improve performance.Minimize the overhead of locking.
Add write capability.

### Impact

faster queries.

iaojnh:
resolve #64

<h3>Greptile Summary</h3>

This PR replaces the old `BufferManager`-based storage backend with a new custom buffer pool (`VecBufferPool`) that maintains a file descriptor and an LRU eviction map, and also fixes context-reset bugs where `reset()` was a no-op in `FlatStreamerContext` and did not clear `group_results_` in `HnswContext`.

**Key changes:**
- New `VecBufferPool` + `LPMap` + `LRUCache` subsystem with per-block mutexes and lock-free concurrent queues for managing a fixed pool of read buffers over an open file descriptor.
- `MemoryBlock` in `index_storage.h` reworked to store a raw `VecBufferPoolHandle*` plus a block ID, with manual ref-counting via `acquire_one`/`release_one`.
- `BufferStorage::open()` now parses the file with `get_meta()` before initialising the pool; `buffer_size_` is read from `BUFFER_STORAGE_MEMORY_SIZE` param (default 2 GiB).
- `context->reset()` is now called on all exit paths in `Index::Add` and `Index::Search`, and `FlatStreamerContext::reset()` / `HnswContext::reset()` actually clear accumulated results.
- `Neighbors` constructor changed from move to copy semantics for `MemoryBlock`, correctly incrementing the ref-count.

**Issues found:**
- `return -1` in `size_t read(…, MemoryBlock &, …)` (line 137) wraps to `SIZE_MAX` — callers cannot detect the error and may interpret it as a successful enormous read.
- `operator=(const MemoryBlock &)` and `operator=(MemoryBlock &&)` both fall through to `default: break` when the source is `MBT_UNKNOWN`; if `this` currently holds an `MBT_BUFFERPOOL` reference, `release_one` is never called and the ref-count is permanently leaked.
- `MemoryBlock` stores a raw (non-owning) `VecBufferPoolHandle*`; if `close_index()` destroys `buffer_pool_handle_` while any `MemoryBlock` (e.g. inside a `Neighbors` struct) is still alive, the destructor dereferences a dangling pointer.
- `VecBufferPool` members `fd_`, `pool_capacity_`, and `no_lru_mode_` lack default initializers — adding them would prevent UB if a code path ever reads them before `init()` completes.

<h3>Confidence Score: 1/5</h3>

- Not safe to merge — multiple correctness bugs in the new MemoryBlock ref-counting and the SIZE_MAX return value sentinel.
- The PR contains two independently reproducible correctness bugs: (1) `return -1` from a `size_t` function silently returns SIZE_MAX, misleading every caller; (2) the copy/move-assignment operator on `MemoryBlock` leaks an LRU ref-count when the source is a moved-from block. Additionally, the raw pointer lifetime issue between `MemoryBlock` and `VecBufferPoolHandle` is an exploitable use-after-free under concurrent close/read scenarios, and previous review rounds already identified several more unresolved issues (memory leaks in init on failure, race conditions in recycle, hardcoded pool size).
- Pay close attention to `src/include/zvec/core/framework/index_storage.h` (MemoryBlock ref-counting and dangling pointer), `src/core/utility/buffer_storage.cc` (SIZE_MAX return value), and `src/ailego/buffer/buffer_pool.cc` (concurrent correctness).

| Filename | Overview |
|----------|----------|
| src/ailego/buffer/buffer_pool.cc | New file implementing LRU-backed buffer pool; has per-block mutex but `no_lru_mode_` and `pool_capacity_` are uninitialized before `init()`, and the `set_block_acquired` / `recycle` path was flagged in prior review for race and memory-leak issues. |
| src/include/zvec/ailego/buffer/buffer_pool.h | New header for VecBufferPool/VecBufferPoolHandle; `fd_`, `pool_capacity_`, and `no_lru_mode_` lack default initializers; destructor correctly frees pooled buffers. |
| src/core/utility/buffer_storage.cc | Major rewrite switching from BufferManager to VecBufferPool; `return -1` from a `size_t` function (line 137) silently returns SIZE_MAX; commented-out debug code left in `open()`; otherwise parse/init logic looks correct. |
| src/include/zvec/core/framework/index_storage.h | MemoryBlock revised to use raw VecBufferPoolHandle* with manual ref-counting; copy/move assignment `default` branch leaks MBT_BUFFERPOOL reference; raw pointer can dangle after close_index(). |
| src/core/algorithm/hnsw/hnsw_entity.h | Neighbors constructor changed from move to copy semantics for MemoryBlock; correctly triggers acquire_one in the copy constructor. |
| src/core/interface/index.cc | context->reset() is now called on all code paths after Add/Search; all early-return paths correctly reset the context before returning. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Caller
    participant BufferStorage
    participant VecBufferPool
    participant LPMap
    participant LRUCache
    participant FD as File (pread)

    Caller->>BufferStorage: open(path)
    BufferStorage->>VecBufferPool: new VecBufferPool(path)
    VecBufferPool->>FD: open() + fstat()
    BufferStorage->>VecBufferPool: init(buffer_size_, max_segment_size_, N)
    VecBufferPool->>LPMap: init(N+10)
    VecBufferPool->>VecBufferPool: ailego_malloc() × buffer_num → free_buffers_

    Caller->>BufferStorage: get(segment_id)
    BufferStorage-->>Caller: WrappedSegment

    Caller->>BufferStorage: segment.read(offset, MemoryBlock, len)
    BufferStorage->>VecBufferPool: acquire_buffer(block_id, offset, capacity, retry=5)
    VecBufferPool->>LPMap: acquire_block(block_id, lru_mode)
    alt block already loaded
        LPMap-->>VecBufferPool: char* (increments ref_count)
    else block not loaded
        VecBufferPool->>VecBufferPool: lock mutex_vec_[block_id]
        VecBufferPool->>VecBufferPool: free_buffers_.try_dequeue(buffer)
        alt no free buffer and LRU mode
            VecBufferPool->>LPMap: recycle(free_buffers_)
            LPMap->>LRUCache: evict_single_block()
            LPMap->>LPMap: evict_block() → returns char* to free_buffers_
        end
        VecBufferPool->>FD: pread(fd_, buffer, capacity, offset)
        VecBufferPool->>LPMap: set_block_acquired(block_id, buffer)
    end
    VecBufferPool-->>BufferStorage: char* raw
    BufferStorage->>BufferStorage: MemoryBlock.reset(handle, block_id, raw+offset)

    Note over BufferStorage: MemoryBlock destructor calls<br/>handle->release_one(block_id)
    BufferStorage->>LPMap: release_block(block_id) [decrements ref_count]
    alt ref_count drops to 0
        LPMap->>LRUCache: add_single_block() [enqueue for future eviction]
    end
```

<details open><summary><h3>Comments Outside Diff (15)</h3></summary>

1. `src/core/utility/buffer_storage.cc`, line 75-93 ([link](https://github.com/alibaba/zvec/blob/9af377ff87d3ed773de080034a3ff2a90c09498e/src/core/utility/buffer_storage.cc#L75-L93)) 

   **Reference count leaked on every `fetch()` call**

   `get_buffer()` calls `acquire_buffer()` → `acquire_block()`, which unconditionally performs a `ref_count.fetch_add(1)`. The `fetch()` function copies the data and returns, but **never calls `release_block()`** to decrement the reference. Every call to `fetch()` permanently increments the block's reference count by one.

   Consequences:
   1. The block is never eligible for eviction (`release_block` adds it to the LRU only when ref_count drops to 0).
   2. On a hot block, `ref_count` will grow unboundedly until signed-integer overflow, which is undefined behaviour in C++.
   3. No buffers are returned to `free_buffers_`, so the pool gradually starves.

   The fix is to call `release_block` after copying the data. The simplest approach is to mirror the RAII pattern already used in `read(MemoryBlock &)`:

   ```cpp
   size_t fetch(size_t offset, void *buf, size_t len) const override {
     // ... bounds check ...
     size_t buffer_offset = ...;
     MemoryBlock block;
     size_t result = const_cast<WrappedSegment*>(this)->read(0, block, capacity_);
     if (!result) return 0;
     auto *data = static_cast<const char*>(block.data()) + offset;
     memmove(buf, data, len);
     return len;
   }
   ```

   Or use an explicit scope guard that calls `release_block`. The same problem affects `read(size_t, const void**, size_t)` (line 96–113): it also acquires a reference via `get_buffer()` and returns a raw pointer without ever releasing the reference.

2. `src/core/utility/buffer_storage.cc`, line 96-113 ([link](https://github.com/alibaba/zvec/blob/9af377ff87d3ed773de080034a3ff2a90c09498e/src/core/utility/buffer_storage.cc#L96-L113)) 

   **Reference count leaked in `read(void**)` path**

   Same issue as `fetch()`: `get_buffer()` increments the block's reference count through `acquire_block()`, but this `read` overload returns a raw `*data` pointer and never calls `release_block()`.

   The pointer returned to the caller remains valid indefinitely (the block cannot be evicted because ref_count ≥ 1 forever), which accidentally provides safety for the raw pointer access. However:
   1. Every call permanently consumes one "slot" in the block's ref_count.
   2. Frequent reads of the same block will cause ref_count to grow without bound, eventually causing signed integer overflow (UB).
   3. Buffer pool resources are never reclaimed.

   ```cpp
   // After this line, raw's block has ref_count incremented:
   auto *raw = owner_->get_buffer(buffer_offset, capacity_, segment_id_);
   if (!raw) {
     return 0;
   }
   *data = raw + offset;
   return len;  // ref_count is leaked — release_block is never called
   ```

   Consider using the `MemoryBlock`-based overload internally, or add an explicit `release_block` call after using the pointer.

3. `src/include/zvec/core/framework/index_storage.h`, line 54-57 ([link](https://github.com/alibaba/zvec/blob/9af377ff87d3ed773de080034a3ff2a90c09498e/src/include/zvec/core/framework/index_storage.h#L54-L57)) 

   **`acquire_one` ignores its return value — silent use-after-free risk in copy assignment**

   `VecBufferPoolHandle::acquire_one()` forwards to `LPMap::acquire_block()`, which returns `nullptr` if the block has been concurrently evicted (i.e. its `ref_count` has been set to `INT_MIN`). `acquire_one` discards this return value entirely:

   ```cpp
   void VecBufferPoolHandle::acquire_one(block_id_t block_id) {
     pool_.lp_map_.acquire_block(block_id);  // return value silently dropped
   }
   ```

   This is exercised in both the copy constructor (line 56) and copy-assignment operator (line 88) of `MemoryBlock`. In the copy-assignment path, `reset()` releases the **old** reference first (decrementing ref_count), potentially allowing another thread to evict the block between the `release_one` and `acquire_one` calls. If the eviction succeeds, `acquire_block` returns `nullptr` — but `acquire_one` ignores this. The `MemoryBlock` now stores a pointer to a buffer that may have been recycled and overwritten, causing a silent use-after-free when `data()` is dereferenced.

   At minimum, `acquire_one` should check the return value and propagate failure to the caller.

4. `src/core/utility/buffer_storage.cc`, line 199-202 ([link](https://github.com/alibaba/zvec/blob/9af377ff87d3ed773de080034a3ff2a90c09498e/src/core/utility/buffer_storage.cc#L199-L202)) 

   **`max_segment_size_` of zero causes `buffer_pool_->init()` to allocate only 10 buffers**

   `max_segment_size_` is computed inside `ParseSegment`. If the index file contains no segments (e.g. a freshly-created index), `max_segment_size_` remains `0` and is passed as `block_size` to `buffer_pool_->init()`:

   ```cpp
   ret = buffer_pool_->init(buffer_size_, max_segment_size_, segments_.size());
   ```

   Inside `init()`:
   ```cpp
   if (block_size == 0) {
     LOG_ERROR("block_size must not be 0");
     return -1;
   }
   ```

   This causes `open()` to return a failure for any empty index, making it impossible to open a freshly-initialised index via `BufferStorage`. This was not a problem with the previous `MMapFile` storage path because it did not require a pre-computed block size at open time.

5. `src/core/utility/buffer_storage.cc`, line 75-82 ([link](https://github.com/alibaba/zvec/blob/d7e60561b54feda958303dd701be10c0dd3e0deb/src/core/utility/buffer_storage.cc#L75-L82)) 

   **`fetch()` is a non-functional stub — silently discards all data**

   `fetch()` has been replaced with `return 0`, meaning it never copies anything into `buf`. This is a critical regression: `memory_read_storage.cc:78` calls `block_->fetch(...)` expecting data to be written into `buf`, and will now silently receive zero bytes. Any code path that reads segment data via `fetch()` will operate on uninitialised memory.

   The full implementation that bounds-checks `offset + len` and reads from the buffer pool (analogous to the `read(MemoryBlock&...)` overload below) must be restored.

6. `src/core/utility/buffer_storage.cc`, line 80-83 ([link](https://github.com/alibaba/zvec/blob/d7e60561b54feda958303dd701be10c0dd3e0deb/src/core/utility/buffer_storage.cc#L80-L83)) 

   **`read(size_t, const void**, size_t)` is a non-functional stub — callers receive stale/null pointers**

   This overload now unconditionally returns 0 without writing to `*data`. Every call site that uses the pointer-return variant is now broken:

   - `src/core/algorithm/flat/flat_searcher.cc:52,142` — `mapping` / `keys_` pointers left uninitialised
   - `src/core/algorithm/hnsw/hnsw_searcher_entity.cc:247,259` — `data` left uninitialised, leading to UB on dereference
   - `src/core/algorithm/hnsw_sparse/hnsw_sparse_searcher_entity.cc:258,270` — same issue
   - `src/core/framework/index_helper.cc:83` — version data will be garbage

   All of these callers check the return value (`!= len`) and proceed to dereference `*data` when they believe the read succeeded. Since this stub always returns 0 (≠ `len`), the callers will at minimum log an error and bail — but any caller that does not strictly check might proceed with a stale or null pointer, causing undefined behaviour.

   A concrete implementation that reads from the buffer pool — mirroring the `read(MemoryBlock&...)` overload — must be provided, or callers must be migrated to the `MemoryBlock` variant.

7. `src/include/zvec/core/framework/index_storage.h`, line 154-155 ([link](https://github.com/alibaba/zvec/blob/d7e60561b54feda958303dd701be10c0dd3e0deb/src/include/zvec/core/framework/index_storage.h#L154-L155)) 

   **Dangling pointer when `close_index()` races with live `MemoryBlock` destructors**

   `buffer_pool_handle_` is stored as a raw non-owning pointer. It points to the `VecBufferPoolHandle` object whose lifetime is controlled by `std::shared_ptr<VecBufferPoolHandle> buffer_pool_handle_` in `BufferStorage`.

   When `close_index()` is called (e.g., from `close()` or the destructor), it calls `buffer_pool_handle_.reset()`, which destroys the `VecBufferPoolHandle`. If any `MemoryBlock` instance (e.g., inside a `Neighbors` struct still held by a caller) outlives that reset, its stored `buffer_pool_handle_` raw pointer is now dangling. The `MemoryBlock` destructor then calls `buffer_pool_handle_->release_one(buffer_block_id_)` on that dangling pointer — undefined behaviour that may crash or silently corrupt memory.

   Consider having `MemoryBlock` hold a `std::shared_ptr<VecBufferPoolHandle>` instead of a raw pointer to prevent the handle from being destroyed while any block is still pinned.

8. `src/core/utility/buffer_storage.cc`, line 164-174 ([link](https://github.com/alibaba/zvec/blob/d7e60561b54feda958303dd701be10c0dd3e0deb/src/core/utility/buffer_storage.cc#L164-L174)) 

   **`buffer_pool_->init()` called with `max_segment_size_` which can be 0**

   `max_segment_size_` is only updated inside the `ParseSegment` loop, so it remains `0` if the index file contains no data-carrying segments (e.g., a freshly created index with only the version segment, or an empty index). Passing `0` as `block_size` to `VecBufferPool::init()` hits the guard `if (block_size == 0) { return -1; }`, causing `open()` to return a hard error and making the index unopenable.

   Additionally, even for valid indices, the `max_segment_size_` is only the size of the **largest** segment. Segments smaller than the max will waste an entire `max_segment_size_`-sized buffer slot per read. Consider validating `max_segment_size_ > 0` before calling `init()` and returning a clear error if the file has no usable segments.

9. `src/core/utility/buffer_storage.cc`, line 199-201 ([link](https://github.com/alibaba/zvec/blob/4231a53f3de66b607ea0d6c737df720555ec1cf8/src/core/utility/buffer_storage.cc#L199-L201)) 

   **`open()` fails for empty or newly-created indices**

   `max_segment_size_` is initialized to `0` and is only updated inside the `for` loop in `ParseSegment()`. If the index file contains no segments (e.g., a freshly written index before any data is added), `max_segment_size_` remains `0` after `ParseToMapping()` returns. This causes `buffer_pool_->init()` to be called with `block_size == 0`, which hits the explicit guard at `buffer_pool.cc:187` and returns `-1`, making `open()` fail entirely.

   A guard (or a fallback minimum block size) is needed before calling `init()`:

   ```cpp
   if (max_segment_size_ == 0) {
     // No segments yet; pool will be re-initialized on first open with data.
     return 0;
   }
   ret = buffer_pool_->init(buffer_size_, max_segment_size_, segments_.size());
   ```

   Alternatively, `init()` could fall back to a sensible minimum block size instead of hard-failing when `block_size == 0`.

10. `src/core/utility/buffer_storage.cc`, line 87-89 ([link](https://github.com/alibaba/zvec/blob/4231a53f3de66b607ea0d6c737df720555ec1cf8/src/core/utility/buffer_storage.cc#L87-L89)) 

    **`fetch()` and `read(const void**)` silently return `0` on buffer pool failure**

    When `get_buffer()` returns `nullptr` (pool exhausted or I/O error), both `fetch()` (line 88) and `read(const void**, size_t)` (line 109) return `0`. Since these methods have a `size_t` return type, callers that use the pattern `if (returned_len != requested_len)` to detect errors will see `0 != len` and may interpret this as a partial/truncated read rather than a hard I/O error — potentially producing silent data corruption downstream.

    The same `return 0` pattern also appears at line 109 in `read(size_t, const void**, size_t)`.

    These methods should propagate a clear failure signal. For `fetch()`, returning `0` can be misread as "zero bytes available". Consider returning a sentinel like `(size_t)-1` on error, consistent with POSIX `read()` semantics, or changing the return type to `int`/`ssize_t` for error-aware callers. At minimum, add a `LOG_ERROR` at these early-exit paths so the failure is visible.

11. `src/core/utility/buffer_storage.cc`, line 131 ([link](https://github.com/alibaba/zvec/blob/4231a53f3de66b607ea0d6c737df720555ec1cf8/src/core/utility/buffer_storage.cc#L131)) 

    **`MemoryBlock` stores a dangling raw pointer when storage is closed**

    `owner_->buffer_pool_handle_.get()` returns a raw `VecBufferPoolHandle*` from the `shared_ptr`. The `MemoryBlock` stores this raw pointer and calls `buffer_pool_handle_->release_one(buffer_block_id_)` in its destructor.

    In `close_index()` (line 468), `buffer_pool_handle_.reset()` destroys the `VecBufferPoolHandle` object. If any `MemoryBlock` returned by `read()` is still alive at that point (e.g., held by a caller), the raw pointer stored in the `MemoryBlock` becomes dangling. When the `MemoryBlock` destructor subsequently calls `release_one()`, this is undefined behaviour — a use-after-free.

    To fix this, `MemoryBlock` should share ownership of the handle's lifetime. For example, store a `shared_ptr<VecBufferPoolHandle>` instead of a raw pointer, and adjust the `MemoryBlock` constructors and `reset()` overloads accordingly. This guarantees the handle stays alive at least as long as any `MemoryBlock` referencing it.

12. `src/core/utility/buffer_storage.cc`, line 137 ([link](https://github.com/alibaba/zvec/blob/d54475261dc04f1dc5cd7e29040f4a4666ababfe/src/core/utility/buffer_storage.cc#L137)) 

    **`return -1` from a `size_t` returning function returns `SIZE_MAX`**

    The return type of `read(size_t offset, MemoryBlock &data, size_t len)` is `size_t` (unsigned). Returning `-1` from an unsigned function does not produce a negative sentinel — it silently wraps to `SIZE_MAX` (the largest possible `size_t` value). Any caller checking `if (ret < 0)` would never catch the error; worse, a caller using the return value as a byte count would behave as if an enormous successful read occurred.

    This branch is only reached when `data.data()` is null *after* `raw` was already confirmed non-null, so it should be unreachable in practice — but the current code compiles with well-defined wrapping semantics and would corrupt every caller that relies on the return value.

    

    Alternatively, change the return type of this override (and the virtual base) to `ssize_t` so that `-1` is representable, or document and consistently use `0` as the error sentinel here.

13. `src/include/zvec/core/framework/index_storage.h`, line 79-95 ([link](https://github.com/alibaba/zvec/blob/d54475261dc04f1dc5cd7e29040f4a4666ababfe/src/include/zvec/core/framework/index_storage.h#L79-L95)) 

    **`operator=` from `MBT_UNKNOWN` source leaks the existing `MBT_BUFFERPOOL` reference**

    When `rhs.type_` is `MBT_UNKNOWN` (e.g., after a move-from), both `operator=(const MemoryBlock &)` (line 79) and `operator=(MemoryBlock &&)` (line 97) hit the `default: break` branch and silently do nothing. If `this->type_` is currently `MBT_BUFFERPOOL`, its reference is never released via `release_one`, and the `lp_map_` ref-count is permanently incremented — leaking a pinned block slot.

    The `reset(void *data)` overload already releases any prior `MBT_BUFFERPOOL` reference, so both assignment operators should call it (with `nullptr`) when `rhs` is `MBT_UNKNOWN`:

    ```cpp
    MemoryBlock &operator=(const MemoryBlock &rhs) {
      if (this != &rhs) {
        switch (rhs.type_) {
          case MemoryBlockType::MBT_MMAP:
            this->reset(rhs.data_);
            break;
          case MemoryBlockType::MBT_BUFFERPOOL:
            this->reset(rhs.buffer_pool_handle_, rhs.buffer_block_id_, rhs.data_);
            buffer_pool_handle_->acquire_one(buffer_block_id_);
            break;
          default:  // MBT_UNKNOWN: release any held reference and become unknown
            if (type_ == MemoryBlockType::MBT_BUFFERPOOL && buffer_pool_handle_) {
              buffer_pool_handle_->release_one(buffer_block_id_);
              buffer_pool_handle_ = nullptr;
            }
            type_ = MemoryBlockType::MBT_UNKNOWN;
            data_ = nullptr;
            break;
        }
      }
      return *this;
    }
    ```

    The same fix is needed for `operator=(MemoryBlock &&)` at line 97.

14. `src/include/zvec/core/framework/index_storage.h`, line 155-158 ([link](https://github.com/alibaba/zvec/blob/d54475261dc04f1dc5cd7e29040f4a4666ababfe/src/include/zvec/core/framework/index_storage.h#L155-L158)) 

    **Raw `VecBufferPoolHandle*` pointer can become a dangling pointer after `close_index()`**

    `MemoryBlock` stores a raw, non-owning `VecBufferPoolHandle *buffer_pool_handle_` (line 157). The owning `shared_ptr<VecBufferPoolHandle>` lives in `BufferStorage::buffer_pool_handle_` and is destroyed when `close_index()` is called (e.g., via `Close()`, `cleanup()`, or the destructor).

    If any `MemoryBlock` object is still alive after `close_index()` — for example, a `Neighbors` struct (see `hnsw_entity.h`) which copies a `MemoryBlock` — calling its destructor will invoke `release_one` through the now-dangling raw pointer, producing undefined behaviour (likely a crash).

    Consider storing a `weak_ptr<VecBufferPoolHandle>` (or a `shared_ptr` of a handle wrapper) inside `MemoryBlock` so that the lifetime relationship is explicit and safe. Alternatively, ensure all callers holding `MemoryBlock` objects are joined/destroyed before `close_index()` is permitted to run, and add a comment or assertion to enforce this invariant.

15. `src/include/zvec/ailego/buffer/buffer_pool.h`, line 130-134 ([link](https://github.com/alibaba/zvec/blob/d54475261dc04f1dc5cd7e29040f4a4666ababfe/src/include/zvec/ailego/buffer/buffer_pool.h#L130-L134)) 

    **Uninitialized private members `no_lru_mode_`, `pool_capacity_`, and `fd_`**

    `fd_`, `pool_capacity_`, and `no_lru_mode_` have no default-member-initializers. While `fd_` is always set in the constructor and `pool_capacity_`/`no_lru_mode_` are set in `init()`, a future code path that reads them before `init()` completes (e.g., if an exception is thrown mid-`open()` and a handle was already obtained) would observe indeterminate values.

</details>

<sub>Last reviewed commit: ["rm posix_fadvise"](https://github.com/alibaba/zvec/commit/d54475261dc04f1dc5cd7e29040f4a4666ababfe)</sub>

### DIFF
--- a/src/ailego/buffer/buffer_pool.cc
+++ b/src/ailego/buffer/buffer_pool.cc
@@ -177,7 +177,6 @@ VecBufferPool::VecBufferPool(const std::string &filename) {
     throw std::runtime_error("Failed to stat file: " + filename);
   }
   file_size_ = st.st_size;
-  posix_fadvise(fd_, 0, 0, POSIX_FADV_RANDOM);
 }
 
 int VecBufferPool::init(size_t pool_capacity, size_t block_size,


Cuiyus:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

src/include/zvec/ailego/buffer/buffer_pool.h
```diff
@@ -0,0 +1,161 @@
...
+  char *get_block(size_t offset, size_t size, size_t block_id);
+
+  int get_meta(size_t offset, size_t length, char *buffer);
+
+  void release_one(block_id_t block_id);
+
+  void acquire_one(block_id_t block_id);
+
+ private:
+  VecBufferPool &pool;
```
richyreachy: use pool_ as member variable

src/include/zvec/ailego/buffer/buffer_pool.h
```diff
@@ -0,0 +1,161 @@
...
+
+ private:
+  std::mutex mutex_;
+  moodycamel::ConcurrentQueue<char *> free_buffers_;
+};
+
+class VecBufferPoolHandle {
+ public:
+  VecBufferPoolHandle(VecBufferPool &pool) : pool(pool) {}
+  VecBufferPoolHandle(VecBufferPoolHandle &&other) : pool(other.pool) {}
```
richyreachy: use std::move to avoid copy

src/core/utility/buffer_storage.cc
```diff
@@ -460,22 +497,22 @@ class BufferStorage : public IndexStorage {
...
   bool index_dirty_{false};
-  mutable IndexMapping mapping_{};
   mutable std::mutex mapping_mutex_{};
 
   // buffer manager
   std::string file_name_;
   IndexFormat::MetaHeader header_{};
   IndexFormat::MetaFooter footer_{};
   std::map<std::string, IndexMapping::SegmentInfo> segments_{};
+  std::map<std::string, size_t> id_hash_{};
```
richyreachy: try unordered hashmap?

src/core/utility/buffer_storage.cc
```diff
@@ -208,14 +233,17 @@ class BufferStorage : public IndexStorage {
     return 0;
   }
 
-  int ParseFooter(int offset) {
-    ailego::BufferHandle footer_handle =
-        get_buffer_handle(offset, sizeof(footer_));
-    void *buffer = footer_handle.pin_vector_data();
+  int ParseFooter(size_t offset) {
+    char *buffer = new char[sizeof(footer_)];
```
richyreachy: std::unique_ptr<char[]> buffer(new char[sizeof(footer_)]);

src/core/utility/buffer_storage.cc
```diff
@@ -174,28 +189,38 @@ class BufferStorage : public IndexStorage {
...
+  int get_meta(size_t offset, size_t length, char *out) {
+    return buffer_pool_handle_->get_meta(offset, length, out);
   }
 
-  int ParseHeader(int offset) {
-    ailego::BufferHandle header_handle =
-        get_buffer_handle(offset, sizeof(header_));
-    void *buffer = header_handle.pin_vector_data();
+  int ParseHeader(size_t offset) {
+    char *buffer = new char[sizeof(header_)];
```
richyreachy: std::unique_ptr<char[]> buffer(new char[sizeof(footer_)]);

src/ailego/buffer/buffer_pool.cc
```diff
@@ -0,0 +1,249 @@
...
+    return buffer;
+  } else {
+    return nullptr;
+  }
+}
+
+char *LPMap::set_block_acquired(block_id_t block_id, char *buffer) {
+  assert(block_id < entry_num_);
+  Entry &entry = entries_[block_id];
+  if (entry.ref_count.load(std::memory_order_relaxed) >= 0) {
```
richyreachy: judge the return value of fetch_add?

src/include/zvec/ailego/buffer/buffer_pool.h
iaojnh: pool_ is a reference member.

src/core/utility/buffer_storage.cc
iaojnh: done

src/core/utility/buffer_storage.cc
iaojnh: done

src/core/utility/buffer_storage.cc
iaojnh: done

src/include/zvec/ailego/buffer/buffer_pool.h
iaojnh: done

src/ailego/buffer/buffer_pool.cc
iaojnh: done

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

iaojnh:
@greptile

PR #87, opened 2026-02-10
fix: fix combined indexer use key instead of index

chinaux:


feihongxu0824:
Please add some unit tests to ensure the fix is reliable.

tests/db/index/segment/segment_test.cc
```diff
@@ -1170,6 +1170,66 @@ TEST_P(SegmentTest, CombinedVectorColumnIndexerWithQuantVectorIndex) {
   ASSERT_EQ(count, 10);
 }
 
+TEST_P(SegmentTest, CombinedVectorColumnIndexerQueryWithPks) {
```
iaojnh: This unit test needs to consider the case of multiple flat indexes.

Issue #92, opened 2026-02-10
[Bug]: SIGILL: Illegal instruction (core dumped) on Intel Broadwell (i5-5200U)


GracePeterMutiibwa:
### Description

I am encountering an `Illegal instruction (core dumped)` error immediately upon trying to create/open a collection. This occurs with `VECTOR_FP32`, `VECTOR_FP16`, and `VECTOR_FP64`.

It appears the PyPI binary wheels may be compiled with AVX-512 instructions enabled, which causes a crash on older architecture CPUs (like Intel Broadwell) that support AVX2 but lack AVX-512.

# Crash occurs here
collection = zvec.create_and_open(path="./zvec_example", schema=schema)

<img width="1289" height="644" alt="Image" src="https://github.com/user-attachments/assets/b4994724-3ae9-41fd-be24-f16b7da231f5" />

# Expected Behavior
The library should either:

Detect the lack of AVX-512 support at runtime and fall back to AVX2 kernels.

Or provide a separate non-AVX-512 build on PyPI for compatibility with older x86_64 processors.

### Steps to Reproduce

```python
import zvec

# Define collection schema
schema = zvec.CollectionSchema(
    name="example",
    vectors=zvec.VectorSchema("embedding", zvec.DataType.VECTOR_FP32, 4),
)
```

### Logs / Stack Trace

```shell
My CPU supports avx and avx2 but does not support avx512. Output of cat /proc/cpuinfo | grep flags | head -1:

flags : ... sse4_1 sse4_2 ... avx ... avx2 ... (avx512 is missing)
```

### Operating System

ubuntu 24.04 (Kernel 6.8.0), Intel Core i5-5200U @ 2.20GHz (Broadwell)

### Build & Runtime Environment

Python 3.12.3, zvec version 0.1.1

### Additional Context

- [ ] I've checked `git status` — no uncommitted submodule changes
- [ ] I built with `CMAKE_BUILD_TYPE=Debug`
- [ ] This occurs with or without `COVERAGE=ON`
- [ ] The issue involves Python ↔ C++ integration (pybind11)

richyreachy:
Hi, could you please attach the backtrace of core dump to help us figure the reason out? To make more clear, can use 'gdb x/10i $pc-20' to show details about the unsupported instruction. Thanks very much.

GracePeterMutiibwa:
## Back Trace Logs 
```bash
(gdb) run
[...]

PR #101, opened 2026-02-13
Add cmake flag for enabling instruction subsets supported by the local machine 

wirthual:
This helped me to work around #92 

To build follow https://zvec.org/en/docs/build/

When using cmake simply set the flag:
```bash
cmake .. -DX86_ARCH_NATIVE=ON
```


feihongxu0824:
Thanks for your contribution! A similar fix has already been implemented in PR #184 , so we’ll close this PR for now.

PR #102, opened 2026-02-13
fix: add cibuildwheel arch overrides to avoid host-native AVX-512 compilation

XBeg9:
## Summary

- Add cibuildwheel overrides to force specific architecture targets for manylinux builds
- x86_64: Enable Haswell target (`-DENABLE_HASWELL=ON`)
- aarch64: Enable ARMv8A target (`-DENABLE_ARMV8A=ON`)

This prevents the build from using host-native AVX-512 instructions that may not be available on all target machines.

## Test plan

- [ ] Verify CI builds pass for manylinux x86_64
- [ ] Verify CI builds pass for manylinux aarch64

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -174,6 +174,15 @@ manylinux-aarch64-image = "manylinux_2_28"
 # Skip 32-bit builds and musllinux
 skip = ["*-manylinux_i686", "*-musllinux*"]
 
+# Force specific architectures to avoid host-native AVX-512 compilation
+[[tool.cibuildwheel.overrides]]
+select = "*-manylinux_*_x86_64"
+environment = { CMAKE_ARGS = "-DENABLE_HASWELL=ON" }
+
+[[tool.cibuildwheel.overrides]]
+select = "*-manylinux_*_aarch64"
+environment = { CMAKE_ARGS = "-DENABLE_ARMV8A=ON" }
+
 [tool.cibuildwheel.macos]
 archs = ["arm64"]
 environment = { MACOSX_DEPLOYMENT_TARGET = "11.0" }


XBeg9:
Related to #101 but takes a different approach:

PR #101 adds a new X86_ARCH_NATIVE CMake flag that enables -march=native builds. This is useful for local development where you want to leverage all CPU features available on your machine. However, it also sets X86_ARCH_NATIVE="ON" as a default in pyproject.toml, which means wheel builds would use the CI machine's native instruction set—potentially including AVX-512 or other instructions not available on end-user machines.

This PR (#102) specifically targets cibuildwheel/CI builds by adding overrides that force safe, widely compatible architecture targets:
- x86_64: Haswell (AVX2, no AVX-512)
- aarch64: ARMv8A (baseline ARM64)

This ensures distributed wheels work on a broad range of hardware without illegal instruction errors (ref #92).

TL;DR: #101 is for local development flexibility; this PR is for ensuring distributed wheels have safe architecture targets in CI.

XBeg9:
Any luck reviewing this, @Cuiyus?

Cuiyus:
> Any luck reviewing this, @Cuiyus?

Hi @XBeg9！ 
We speculate that the "illegal instruction" that occurred in #101 might be caused by the fact that the std::string compare within the std::map set was enabled for vectorization. We are currently trying to fix it. 

Thank you very much for your submission. Here, zvec focuses on addressing specific issues and determining solutions. Before reaching a conclusion on #101, modifications to cibuildwheel will not be made. 

Thank you again for your attention! If you don't mind, I'll first close this PR. The subsequent solutions will be synchronized under #101 .

feihongxu0824:
Thank you for your contribution! A similar feature has already been implemented in PR #184 , so we’ll close this PR for now.

PR #146, opened 2026-02-17
Allow custom model names for Dense, Sparse embeddings, fix Sparse Embeddings

tomaarsen:
Hello!

## Pull Request overview
* Allow custom model_name in `DefaultLocalSparseEmbedding` and `DefaultLocalDenseEmbedding`
* Simplify model loading, only 1 `_get_model` across DenseEmbedding, SparseEmbedding, and ReRanker via a `_get_model_class` property
* Fix SparseEncoder model loading: It used SentenceTransformer previously. Also, `encode_query` & `encode_document` exists for both SparseEncoder and SentenceTransformer.
* Removed `_manual_sparse_encode` for older Sentence Transformer: I would recommend requiring at least Sentence Transformers v5.0 so the SparseEncoder can be used.
* Replaced `SentenceTransformerReRanker` with `DefaultLocalReRanker` in the docs, as the previous doesn't exist anymore.

## Details:
This PR extends the support for local dense, sparse, and reranker models a good bit. I've had some installation issues, so I'm unable to properly run the tests at the moment, but you should get much clearer results for the SparseEmbedding class. I also simplified the sparse embedding post-processing using `model.decode`, which also removes some issues with SparseCUDA and SparseCPU not implementing all operations.

Please do let me know if you have any questions.

- Tom Aarsen

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -52,6 +52,7 @@ test = [
     "pytest-cov >=4.1",
     "pytest-mock >=3.12",
     "cibuildwheel == 2.17.0",
+    "sentence-transformers",
 ]
 docs = [
     "mkdocs >=1.5",


python/zvec/extension/sentence_transformer_embedding_function.py
```diff
@@ -39,6 +40,9 @@ class DefaultLocalDenseEmbedding(
     similarity tasks. It runs locally without requiring API keys.
 
     Args:
+        model_name (Optional[str]): Model identifier or local path. Defaults to:
```
Cuiyus: It is recommended not to expose the model_name in args. We expect users to regard `DefaultLocalDenseEmbedding` as a ready-to-use final class, rather than a tool class that encapsulates the Sentence Transform Loading mode process. 

Since once the `model_name` parameter is exposed, the logic of loading mode and parser output needs to be considered. Therefore, I think the '_get_model_class' method you proposed is a good approach, allowing users to handle the logic of loading the model (including the version requirements for Sentence Transform) by themselves. 

When the user expects to use another local model, they can implement it by inheriting from SentenceTransformerFunctionBase and DenseEmbeddingFunction themselves.

python/zvec/extension/sentence_transformer_function.py
```diff
@@ -88,6 +88,18 @@ def device(self) -> str:
             return str(model.device)
         return self._device or "cpu"
 
+    @property
+    def _get_model_class(self):
```
Cuiyus: Here, it would be more appropriate to replace _get_model_class with _loading_model.

python/zvec/extension/sentence_transformer_embedding_function.py
tomaarsen: I understand, I've reverted the changes in 9abce9fcc0e81f27fc6db27c06afef1940db8301. 
I've not updated the ReRanker, as it already supported the `model_name`.

python/zvec/extension/sentence_transformer_function.py
tomaarsen: What would you like this method to do? Currently in the PR, the `_get_model` method is responsible for actually loading the model, and `_get_model_class` is just a method for choosing which model class to load. This allows the 3 subclasses to only require ~2 lines of code for the model loading, instead of overriding the full model loading (with ModelScope support, try-except ImportError catching, etc.).

tomaarsen:
Thank you for the review! I'd be glad to help move this PR forward together. I fixed one of your concerns, and I had a question for the other one. You're also free to take over the PR if you prefer (I know I sometimes find it easier 😄).

- Tom Aarsen

python/zvec/extension/sentence_transformer_function.py
Cuiyus: Agree with your reply. 
The relevant responsibilities of _get_model_class are exactly as you stated.

python/zvec/extension/sentence_transformer_function.py
```diff
@@ -88,6 +88,18 @@ def device(self) -> str:
             return str(model.device)
         return self._device or "cpu"
 
+    @property
```
Cuiyus: Why is _get_model_class being decorated with @property here (instead of being defined as an abstract method and not exposed externally)? 

Are you hoping to expose the Sentence Transformer class so that users can perform other operations or learn more about Sentence Transformer?

Cuiyus:
@greptile

Cuiyus:
> Thank you for the review! I'd be glad to help move this PR forward together. I fixed one of your concerns, and I had a question for the other one. You're also free to take over the PR if you prefer (I know I sometimes find it easier 😄).
> 
> * Tom Aarsen

Let's work together to merge PR. ❤️ 
I noticed that the ruff check in the CI failed. Please take the time to fix it.

tomaarsen:
I addressed greptile's valid comments and ran `ruff`!

- Tom Aarsen

python/zvec/extension/sentence_transformer_function.py
tomaarsen: Ah, yeah it really doesn't need to be a property. I think I got distracted by the `model_name`, `model_source`, and `device` properties above 😄 I'll turn it into a normal method. Resolved in 2211e4dc70891b85c035cf811ffaa8b79906becf

Cuiyus (approved):

Cuiyus:
@tomaarsen I noticed that the Python unit tests in the ci stage failed to run. After the single test is fixed, I will merge this PR.

tomaarsen:
Apologies, that was an oversight on my side. I forgot to update how `_get_model_class` is being used when I updated it from a property to a method.

- Tom Aarsen

Cuiyus:
> Apologies, that was an oversight on my side. I forgot to update how `_get_model_class` is being used when I updated it from a property to a method.
> 
> * Tom Aarsen

Hi @tomaarsen. The CI still failed to pass. I noticed that some cases in the Python UT have needed to be adjusted.

tomaarsen:
I've added Sentence Transformers to CI installs. I can also add it to pyproject.toml under the test dependency, what do you think? I'm not sure what your preference is.

- Tom Aarsen

Cuiyus:
> I've added Sentence Transformers to CI installs. I can also add it to pyproject.toml under the test dependency, what do you think? I'm not sure what your preference is.
> 
> * Tom Aarsen

It is also advisable to make the update under the test dependency section in pyproject.toml.

tomaarsen:
Done. I believe the `test` extra isn't being installed in the CI, so I added it directly in the CI install earlier.

- Tom Aarsen

tomaarsen:
Hmm, it looks like the existing tests were never run before, they presumably didn't work before either, as `sentence-transformers` was never imported or installed before? 

- Tom Aarsen

Cuiyus:
> Hmm, it looks like the existing tests were never run before, they presumably didn't work before either, as `sentence-transformers` was never imported or installed before?
> 
> * Tom Aarsen

Ohhhh! 
I'm very sorry. Upon reviewing the embedding unit tests, sentence-transformers is mocked, not actually loaded. You can solve this by replacing the line `@patch("zvec.extension.sentence_transformer_function.require_module")` in test_embeding.py with `@patch("zvec.extension.sentence_transformer_embedding_function.require_module")`. (test_rerank.py is the same.) 

Of course, some of the logic in the unit tests may need to be adjusted.

Cuiyus:
Hi~ I'll first close this PR. Once the unit test fixes are approved, you can reopen it. Thank you for your attention to zvec.

PR #149, opened 2026-02-19
feat(examples): add custom HTTP embedding example for LM Studio / Ollama

cluster2600:
## Summary

This PR adds a self-contained example showing how to use any **OpenAI-compatible HTTP embedding endpoint** (LM Studio, Ollama, vLLM, LocalAI, …) as the embedding source in zvec.

## What's added

**`examples/custom_http_embedding.py`**

| | |
|---|---|
| **`HTTPEmbeddingFunction`** | A zero-dependency class (stdlib only) that calls any `/v1/embeddings` endpoint, caches results with `@lru_cache`, and satisfies the `DenseEmbeddingFunction` protocol. |
| **Collection setup** | HNSW index with cosine similarity, dimension auto-detected from the server response. |
| **Insert + query** | 5 sample documents embedded on the fly, then a semantic search query. |
| **CLI interface** | `--base-url`, `--model`, `--api-key`, `--collection-path` flags for easy customisation. |
| **README-style header** | Step-by-step instructions for LM Studio and Ollama embedded at the top of the file. |

## Usage

```bash
# LM Studio (default)
python examples/custom_http_embedding.py

# Ollama
python examples/custom_http_embedding.py \
    --base-url http://localhost:11434 \
    --model nomic-embed-text
```

## Motivation

The existing extensions (`OpenAIDenseEmbedding`, etc.) require the `openai` package and are primarily designed for cloud APIs. Many developers want to use local inference servers without extra dependencies. This example shows the pattern using only Python stdlib, making it easy to adapt or inline.

## Testing

The example runs end-to-end against a live LM Studio instance on `localhost:1234`. No new test infrastructure is required for a standalone script.


Cuiyus:
Thank you for your submission!

This service-oriented model for invocation, which helps zvec achieve the RAG capability, is what we currently lack. 
You can implement your HttpEmbedingFunction(OllmaEmbedingFunction) by inheriting `DenseEmbedingFunction` in the directory python/zvec/extension. This will make it easier for more users to use!

https://zvec.org/api-reference/python/extension/#zvec.extension.DenseEmbeddingFunction

cluster2600:
Thanks for the feedback\! Moved the implementation into `python/zvec/extension/http_embedding_function.py` as `HTTPDenseEmbedding`, inheriting from `DenseEmbeddingFunction`. It's now exported from `zvec.extension` and the example imports it from there instead of defining the class inline.

Cuiyus:
@greptile

examples/custom_http_embedding.py
```diff
@@ -0,0 +1,232 @@
+"""
```
Cuiyus: This "http_embeding_function" demo can be deleted.

These examples that cannot be directly run (and require the llm server to be started) can be migrated to the zvec-web project.

Let users understand how to use it through the instructions provided on the [zvec-web](https://github.com/zvec-ai/zvec-web).

cluster2600:
Removed the example file as requested — server-dependent examples belong in [zvec-web](https://github.com/zvec-ai/zvec-web).

Cuiyus:
> Removed the example file as requested — server-dependent examples belong in [zvec-web](https://github.com/zvec-ai/zvec-web).

I didn't find any pull requests related to HttpEmbedingFunction on the link https://github.com/zvec-ai/zvec-web/pulls.

It would be helpful if you could add some relevant explanations about the HttpEmbedingFunction in the zvec-web section, so that more users can use it.

Cuiyus:
@greptile

PR #152, opened 2026-02-19
chore: avoid redundant IndexDocument construction in heap emplace

longway-code:
## Summary
`IndexDocumentHeap::emplace` now checks `score < front().score()` before calling into `ailego::Heap`, avoiding unnecessary object construction for candidates that would be rejected anyway.

## Performance
Tested locally, decreased ~15% end-to-end latency of `FlatSearcher::search_impl` (N=100k, dim=128, topk=10, FP32 SquaredEuclidean, 200 queries averaged).

All Tests Passsed


iaojnh:
@greptile

richyreachy:
hi @longway-code, thanks for your kindly contribution. 

As we see the fix, the condition to avoid emplace, is quite same as the emplace inner logical condition.  

We'll take a test to see if this can bring the mentioned performance improvement.

richyreachy:
Hi, @longway-code, a test we took by @iaojnh shows there's no remarkable improvement over current codes. Could you please provide more details about the test setup and datasets etc. ?

richyreachy:
I close this for now. We can refer back to it once the related issue is on the table. Thanks.

PR #153, opened 2026-02-20
chore(flat): replace std::map with std::unordered_map for key_id_mapping

longway-code:
`key_id_mapping_` is only used for point lookups — no range scans, no ordering needed. `std::unordered_map` is a better fit.

- `get_id()` lookup: O(log N) → O(1)
- Index load: O(N log N) → O(N)
- Memory: slightly lower — three tree pointers (parent/left/right) per node replaced by one `next` pointer plus a bucket array; net ~12MB savings at N=1M

**Microbenchmark** — isolated `map::find` vs `unordered_map::find`, random key order, all hits:

| N | std::map | std::unordered_map | speedup |
|---|---|---|---|
| 10k | 57.3 ns/op | 7.0 ns/op | 8.2x |
| 100k | 88.3 ns/op | 10.6 ns/op | 8.3x |
| 1M | 138.5 ns/op | 15.1 ns/op | 9.2x |

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+version: 2
+updates:
+  # GitHub Actions dependencies
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "02:00"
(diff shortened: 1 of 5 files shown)


iaojnh:
@greptile

PR #157, opened 2026-02-22
feat: add Python 3.13 and 3.14 support

cluster2600:
## Summary

Add support for Python 3.13 and 3.14 in zvec.

### pyproject.toml
- Add Python 3.13 and 3.14 to classifiers
- Add cp313-* to cibuildwheel build targets
- Update ruff target-version to py313

### CI Workflows
- linux_x64_docker_ci.yml: Add Python 3.13 to test matrix
- linux_arm64_docker_ci.yml: Add Python 3.13 to test matrix
- mac_arm64_ci.yml: Add Python 3.13 to test matrix

## Testing

- [x] CI workflows updated to test Python 3.13
- [x] cibuildwheel configured to build cp313 wheels
- [x] Verify CI builds pass

## Related Issues

Fixes #131

PR #165, opened 2026-02-24
feat: add compression module with zstd/gzip/lzma support

cluster2600:
## Summary
- Add Python compression module with zstd, gzip, and lzma backends
- Add streaming compression API for large datasets
- Enable RocksDB runtime compression codec detection (ZSTD → LZ4 → Snappy fallback)
- Extend `CollectionSchema` with compression configuration

### Python
- `python/zvec/compression.py` — Core compression with pluggable backends
- `python/zvec/compression_integration.py` — Collection-level compression integration
- `python/zvec/streaming.py` — Streaming compression/decompression API
- `python/zvec/model/schema/collection_schema.py` — Compression config in schema
- `python/zvec/__init__.py` — Export compression module

### C++
- `src/db/common/rocbsdb_context.cc` — Runtime detection of supported compression codecs using `rocksdb::GetSupportedCompressions()`, with fallback chain: ZSTD → LZ4 → Snappy → None

### Tests
- `python/tests/test_compression.py`
- `python/tests/test_compression_integration.py`
- `python/tests/test_schema_compression.py`
- `python/tests/test_streaming.py`

### Docs
- `docs/COMPRESSION.md` — Compression usage and configuration guide

## Context
Split from #157 to isolate compression feature from CI and GPU work.

## Test plan
- [ ] ruff lint and format pass
- [ ] clang-format passes on `rocbsdb_context.cc`
- [ ] Compression tests pass locally
- [ ] CI builds succeed (RocksDB compression fallback works on all platforms)

### DIFF
--- a/docs/COMPRESSION.md
+++ b/docs/COMPRESSION.md
@@ -1,0 +1,272 @@
+# Compression Guide
+
+This guide explains how to use zvec's compression features to reduce storage size and improve performance.
+
+## Overview
+
+zvec provides compression at two levels:
+
+1. **Python Level**: Pre/post-processing compression for vectors
+2. **C++ Level**: Automatic RocksDB storage compression
+
+## Installation
+
+Compression features are built-in. For optimal performance with zstd, install Python 3.13+:
+
+```bash
+# Python 3.13+ recommended for zstd support
+pip install zvec
+```
+
+## C++ Storage Compression
+
+The C++ storage layer uses **RocksDB** with automatic compression:
+
+| Level | Compression | Use Case |
+|-------|-------------|----------|
+| 0 (memtable) | None | Speed |
+| 1-2 | LZ4 | Fast warm data |
+| 3-6 | Zstd | Best compression |
+
+This is automatic and transparent - all data stored in zvec collections is compressed.
+
+**Benefits:**
+- No configuration needed
+- Transparent to users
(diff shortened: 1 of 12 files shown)


zhourrr:
Thanks for putting together this compression proposal — we appreciate the effort and thought you've put into the design!

Our system is built around a C++ core, with Python (and other languages) serving as bindings to this vector search engine. Because of this, we generally aim to keep data-handling logic in the C++ engine, so all bindings benefit uniformly and behavior remains consistent. Are there particular use cases or constraints that make compression in Python necessary or advantageous?

I'm also curious about the expected benefit for vector data specifically. High-dimensional floating-point vectors are typically not very compressible with general-purpose algorithms. For storage efficiency, we already support quantizations (though they hurt recall), which are typically more effective for vectors than byte-level compression.

Any additional context you can share about the problem you're trying to solve would be very helpful!

cluster2600:
Thanks for the thoughtful feedback @zhourrr! You raise valid points — keeping data-handling logic in the C++ core makes sense for consistency across all language bindings, and I agree that general-purpose compression has limited benefit for high-dimensional float vectors compared to the quantization you already support.

I'll close this one out. If compression at the C++ engine level ever becomes useful, happy to contribute there instead. Appreciate the review!

PR #191, opened 2026-03-02
fix: remove crash residue

zhourrr:


zhourrr:
Closed as implemented in #208

PR #208, opened 2026-03-09
fix: clean up crash residue

zhourrr:
<h3>Greptile Summary</h3>

This PR introduces crash-residue cleanup logic: before constructing a `VectorColumnIndexer` or `MemForwardStore`, it now detects and removes pre-existing files that are likely left over from a previous crash. It also ships a new `crash_recovery` test suite (a data-generator binary + four `CrashRecoveryTest` cases) and refactors `Doc::operator==` to use epsilon-based float comparison via a new `ValueEqual` visitor.

Key observations:
- The two new `FileHelper::RemoveFile` calls in `segment.cc` do not check the return value, unlike every other call site in the same file — if removal fails silently, the subsequent construction proceeds on top of the corrupted residue file, defeating the purpose of the guard (previously flagged).
- `write_recovery_test.cc` is missing `<unistd.h>` and `<sys/wait.h>` for `fork`/`execvp`/`waitpid`/`pid_t` (previously flagged); `BasicInsertAndReopen` tests a clean write-then-reopen scenario instead of an actual crash-recovery path (previously flagged).
- `SetUp`/`TearDown` use `system("rm -rf ...")` with an ignored return value; `std::filesystem::remove_all()` (already included) is the correct portable replacement.
- `RunGeneratorAndCrash` contains a TOCTOU race between `kill(pid, 0)` and `kill(pid, SIGKILL)` that can cause `ASSERT_TRUE(WIFSIGNALED(status))` to fire spuriously if the child exits naturally in the narrow window between the two calls.
- `data_generator.cc` uses `999999` as a magic sentinel for an unset `--version`; `std::optional<int>` is the idiomatic alternative and avoids silently rejecting a legitimately-supplied value of `999999`.

- Not safe to merge — the core crash-cleanup logic can silently proceed on top of a corrupted residue file, and the test suite has compilation issues and a race condition that make it unreliable.
- The unchecked `RemoveFile` return values in `segment.cc` mean a failed deletion (e.g., permissions, I/O error) goes unnoticed and the code continues constructing an indexer or forward store on the corrupt file — the exact scenario this PR is meant to prevent. Combined with missing POSIX headers that will fail compilation on strict toolchains, a `BasicInsertAndReopen` test that does not exercise the new code paths, and a TOCTOU race in `RunGeneratorAndCrash`, the PR has multiple correctness gaps that should be addressed before merging.
- `src/db/index/segment/segment.cc` (unchecked `RemoveFile` return values) and `tests/db/crash_recovery/write_recovery_test.cc` (missing headers, non-testing crash path, TOCTOU race).

| Filename | Overview |
|----------|----------|
| src/db/index/segment/segment.cc | Adds `FileExists` + `RemoveFile` guards before creating `VectorColumnIndexer` and `MemForwardStore` to clean up crash-residue files; also fixes a comment/ordering swap between UPSERT and UPDATE in the WAL recovery log. Both new `RemoveFile` calls do not check the return value, unlike other callers in the same file (previously flagged). |
| src/db/index/common/doc.cc | Refactors `Doc::operator==` to use a new `ValueEqual` visitor struct that adds float/double epsilon tolerance; the old code used strict `==` for all types. Change is correct and adds `<cmath>` as needed. |
| src/include/zvec/db/doc.h | Forward-declares `Doc::ValueEqual` as a public nested struct to allow its definition in `doc.cc`; minimal, correct change. |
| tests/db/crash_recovery/write_recovery_test.cc | New crash-recovery test suite using `fork`/`execvp`/`kill`; missing POSIX headers (`<unistd.h>`, `<sys/wait.h>`), `BasicInsertAndReopen` does not exercise crash-residue logic, `system("rm -rf ...")` ignores return value and is non-portable, and there is a TOCTOU race in `RunGeneratorAndCrash` between the liveness check and `SIGKILL`. |
| tests/db/crash_recovery/data_generator.cc | Standalone data-generator binary used by the test suite to insert/update/upsert/delete documents in batches; uses magic sentinel `999999` for unset `--version`, preventing valid use of that version number. |
| tests/db/crash_recovery/utility.h | Provides `CreateTestSchema` and `CreateTestDoc` helpers with deterministic, version-keyed document values; no issues found. |
| tests/db/crash_recovery/CMakeLists.txt | New CMake file building `data_generator` and test executables; correctly adds `data_generator` as a dependency of each test target. |
| tests/db/CMakeLists.txt | Adds `cc_directories(crash_recovery)` to include the new test subdirectory; trivial and correct. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[SegmentImpl::create_vector_indexer] --> B{FileHelper::FileExists\nindex_file_path?}
    B -- Yes --> C[LOG_WARN crash residue]
    C --> D[FileHelper::RemoveFile\nreturn value unchecked ⚠️]
    D --> E[make_shared VectorColumnIndexer]
    B -- No --> E

    F[SegmentImpl::init_memory_components] --> G{FileHelper::FileExists\nmem_path?}
    G -- Yes --> H[LOG_WARN crash residue]
    H --> I[FileHelper::RemoveFile\nreturn value unchecked ⚠️]
    I --> J[make_shared MemForwardStore]
    G -- No --> J

    K[CrashRecoveryTest] --> L[SetUp: system rm -rf ⚠️]
    L --> M{Test type}
    M -- BasicInsertAndReopen --> N[RunGenerator — normal run\nno crash simulated ⚠️]
    M -- CrashRecoveryDuringInsertion/Upsert/Update/Delete --> O[RunGeneratorAndCrash\nfork + sleep + kill]
    O --> P{kill PID 0 == 0?\nTOCTOU race ⚠️}
    P -- alive --> Q[kill SIGKILL]
    Q --> R[waitpid — assert WIFSIGNALED]
    P -- already exited --> R
    N --> S[Collection::Open + verify doc count]
    R --> S
```

<details open><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `tests/db/crash_recovery/write_recovery_test.cc`, line 760-766 ([link](https://github.com/alibaba/zvec/blob/a0b9841e99a195246d43d176d6117b8e3e0cd39d/tests/db/crash_recovery/write_recovery_test.cc#L760-L766)) 

   **`system()` should be replaced with `std::filesystem::remove_all()`**

   `system("rm -rf ./crash_test_db")` is used in both `SetUp` and `TearDown`, but its return value is silently ignored and it is non-portable (fails on Windows). Since `<filesystem>` is already included in this file, `std::filesystem::remove_all()` is the idiomatic and safer alternative.

   

   The same applies to `TearDown`:
   ```cpp
     void TearDown() override {
       std::filesystem::remove_all(dir_path_);
     }
   ```

2. `tests/db/crash_recovery/data_generator.cc`, line 283 ([link](https://github.com/alibaba/zvec/blob/a0b9841e99a195246d43d176d6117b8e3e0cd39d/tests/db/crash_recovery/data_generator.cc#L283)) 

   **Magic sentinel value `999999` for unset version**

   Using `999999` as a default sentinel to detect "not supplied by the user" is fragile: a caller passing `--version 999999` will be silently rejected as if the argument were missing, with no helpful error message. `std::optional<int>` is the idiomatic way to represent an optional integer argument.

   

   Then in `ParseArgs`, change the assignment to:
   ```cpp
   config.version = std::stoi(argv[++i]);
   ```
   And the validation check to:
   ```cpp
   if (config.path.empty() || config.operation.empty() ||
       config.start_id >= config.end_id || !config.version.has_value()) {
     return false;
   }
   ```

3. `tests/db/crash_recovery/write_recovery_test.cc`, line 747-753 ([link](https://github.com/alibaba/zvec/blob/a0b9841e99a195246d43d176d6117b8e3e0cd39d/tests/db/crash_recovery/write_recovery_test.cc#L747-L753)) 

   **TOCTOU race between `kill(pid, 0)` and `kill(pid, SIGKILL)`**

   There is a time-of-check/time-of-use race: between `kill(pid, 0) == 0` (which confirms the process is alive) and the subsequent `kill(pid, SIGKILL)`, the child process could complete naturally and exit. If that happens, the process was not killed by a signal, so `WIFSIGNALED(status)` will be false and `ASSERT_TRUE` will fail spuriously.

   A more robust pattern is to send `SIGKILL` unconditionally and accept that the child may have already exited:

   

   Alternatively, document that the test is designed so the crash window is always shorter than the generator's total run time to justify the unconditional `ASSERT_TRUE(WIFSIGNALED(status))`.

</details>

<sub>Last reviewed commit: a0b9841</sub>

### DIFF
--- a/tests/db/crash_recovery/data_generator.cc
+++ b/tests/db/crash_recovery/data_generator.cc
@@ -21,15 +21,16 @@
 #include "utility.h"
 
 
-constexpr uint64_t kBatchSize = 20;
-constexpr uint64_t kBatchDelayMs = 10;
+constexpr int kBatchSize = 20;
+constexpr int kBatchDelayMs = 10;
 
 
 struct Config {
   std::string path;
-  uint64_t start_id = 0;
-  uint64_t end_id = 0;
-  std::string operation;  // "insert", "upsert", "update"
+  int start_id = 0;
+  int end_id = 0;
+  std::string operation;  // "insert", "upsert", "update", "delete"
+  int version = 999999;
 };
 
 
@@ -40,11 +41,13 @@ bool ParseArgs(int argc, char **argv, Config &config) {
     if (arg == "--path" && i + 1 < argc) {
       config.path = argv[++i];
     } else if (arg == "--start" && i + 1 < argc) {
-      config.start_id = std::stoull(argv[++i]);
+      config.start_id = std::stoi(argv[++i]);
     } else if (arg == "--end" && i + 1 < argc) {
-      config.end_id = std::stoull(argv[++i]);
+      config.end_id = std::stoi(argv[++i]);
     } else if (arg == "--op" && i + 1 < argc) {
       config.operation = argv[++i];
+    } else if (arg == "--version" && i + 1 < argc) {
+      config.version = std::stoi(argv[++i]);
     } else if (arg == "--help" || arg == "-h") {
       return false;
     }
@@ -52,15 +55,16 @@ bool ParseArgs(int argc, char **argv, Config &config) {
 
   // Validate required arguments
   if (config.path.empty() || config.operation.empty() ||
-      config.start_id >= config.end_id) {
+      config.start_id >= config.end_id || config.version == 999999) {
     return false;
   }
 
   // Validate operation
   if (config.operation != "insert" && config.operation != "upsert" &&
-      config.operation != "update") {
+      config.operation != "update" && config.operation != "delete") {
     std::cerr << "Error: Invalid operation '" << config.operation
-              << "'. Must be 'insert', 'upsert', or 'update'." << std::endl;
+              << "'. Must be 'insert', 'upsert', 'update', or 'delete'."
+              << std::endl;
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


zhourrr:
@greptile

PR #213, opened 2026-03-11
feat: enable icelake and l2 batch distance for int8 quantization.

luoxiaojian:
<h3>Greptile Summary</h3>

This PR extends INT8 quantization support to Intel IcelAke CPUs by wiring up VNNI-based `InnerProductDistanceBatch` for the `SquaredEuclidean` metric, expanding the per-vector extra storage from 16 → 20 bytes (or 20 → 24 for cosine) to accommodate a new `int8_sum` field used to correct the +128 unsigned bias introduced by VNNI preprocessing.

Key changes:
- **`cmake/option.cmake`**: Adds `ENABLE_ICELAKE` build option and inserts `icelake-server` into the AVX-512 capability probe list.
- **`record_quantizer.h`**: All INT8 vectors now always write both `squared_sum` at `extras[3]` and `int8_sum` at `extras[4]`, regardless of metric. This is a **breaking on-disk format change** (16 → 20 bytes for INT8 extras); existing indexes built with the old format will be silently misread without any version guard.
- **`quantized_integer_metric_batch.h`**: `SquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>` conditionally subtracts `128 * int8_sum` when the IcelAke preprocess function is active; all INT8 batch structs update their `dim - N` offsets accordingly. `MipsSquaredEuclidean` receives the dim-offset fix but is left without `GetQueryPreprocessFunc` or the int8_sum correction — the IcelAke path for that metric is incomplete.
- **`quantized_integer_metric_matrix.h`**: Offsets bumped to match new format, but the `SquaredEuclidean` matrix compute path does not apply the VNNI int8_sum correction — whether this is safe depends on whether `ailego::InnerProductMatrix` uses VNNI on IcelAke.
- **Converters / reformers**: Extra dimension constants updated consistently across `integer_quantizer_converter.cc`, `integer_quantizer_reformer.cc`, `cosine_converter.cc`, and `cosine_reformer.cc`.

- Not safe to merge as-is due to a silent breaking on-disk format change and unresolved correctness questions in the matrix and MIPS batch paths on IcelAke hardware.
- The format change in record_quantizer.h is backward-incompatible with no version guard, the SquaredEuclidean matrix path may produce wrong results on IcelAke if ailego::InnerProductMatrix uses VNNI, and MipsSquaredEuclidean is left in a half-implemented IcelAke state with the int8_sum correction absent from ComputeBatch. Test coverage also doesn't exercise the IcelAke VNNI correction branch.
- Focus on `src/core/metric/quantized_integer_metric_matrix.h` (missing VNNI correction in SquaredEuclidean matrix path), `src/core/metric/quantized_integer_metric_batch.h` (incomplete MipsSquaredEuclidean IcelAke support), and `src/core/quantizer/record_quantizer.h` (breaking format change without version guard).

| Filename | Overview |
|----------|----------|
| src/core/metric/quantized_integer_metric_batch.h | Core batch distance dispatch updated for IcelAke VNNI: SquaredEuclidean now conditionally applies 128*int8_sum correction and gets a QueryPreprocess path; MipsSquaredEuclidean receives the dim offset fix but is left without GetQueryPreprocessFunc or the int8_sum correction, creating a latent correctness gap. |
| src/core/metric/quantized_integer_metric_matrix.h | Extra dimension offsets bumped from 16→20 (INT8 non-cosine) and 20→24 (cosine INT8) to account for the new int8_sum slot, but the SquaredEuclidean matrix compute path never reads or applies the 128*int8_sum VNNI correction that the batch path now uses. |
| src/core/quantizer/record_quantizer.h | Now unconditionally writes both squared_sum (extras[3]) and int8_sum (extras[4]) for all INT8 vectors, expanding per-vector storage from 16 to 20 bytes; INT4 path is left using the non-portable int* cast for its int8_sum write. |
| src/core/metric/quantized_integer_metric.cc | Adds IcelAke query-preprocess dispatch for SquaredEuclidean INT8, mirroring the existing Cosine path; MipsSquaredEuclidean remains unregistered here which is consistent with the batch header leaving it unimplemented. |
| cmake/option.cmake | Adds ENABLE_ICELAKE option and correctly inserts icelake-server into the AVX-512 capability probe list; the manual add_arch_flag block places IcelAke after Sapphire Rapids but before Skylake-AVX512, slightly out of ISA capability order. |
| tests/core/metric/quantized_integer_metric_test.cc | Dimension assertions updated from +16 to +20 for INT8 matrix tests; no test exercises the IcelAke-specific VNNI correction branch (GetQueryPreprocessFunc != nullptr path), leaving the new conditional logic without coverage. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Caller
    participant QIM as QuantizedIntegerMetric
    participant Base as BaseDistanceBatchWithScoreUnquantized
    participant SEB as SquaredEuclideanBatch<int8_t>
    participant IPB as InnerProductDistanceBatch (ailego)
    participant GPF as GetQueryPreprocessFunc (IcelAke)

    Note over Caller,GPF: batch_distance() path (new IcelAke flow)

    Caller->>QIM: get_query_preprocess_func()
    QIM->>SEB: GetQueryPreprocessFunc()
    SEB-->>QIM: &QueryPreprocess (non-null)
    QIM-->>Caller: QueryPreprocess fn ptr

    Caller->>SEB: QueryPreprocess(query, dim)
    SEB->>IPB: GetQueryPreprocessFunc()
    IPB-->>SEB: func (non-null on IcelAke)
    SEB->>func: func(query, dim - 20)
    Note right of func: VNNI preprocess: shift query by +128

    Caller->>QIM: batch_distance()(vecs, query, n, dim, results)
    QIM->>Base: ComputeBatch(vecs, query, n, dim, results)
    Base->>SEB: ComputeBatch(vecs, query, n, dim, results)
    SEB->>IPB: ComputeBatch(vecs, query, n, original_dim, results)
    Note right of IPB: Raw VNNI inner products (biased by 128*int8_sum)
    IPB-->>SEB: raw results[]
    SEB->>IPB: GetQueryPreprocessFunc() != nullptr?
    alt IcelAke hardware
        SEB->>SEB: result -= 128 * int8_sum  (correction)
    end
    SEB->>SEB: Apply SquaredEuclidean formula
    SEB-->>Caller: corrected distances
```

1. `src/core/metric/quantized_integer_metric_batch.h`, line 293-320 ([link](https://github.com/alibaba/zvec/blob/7dfac9c26fbcf969921283466addd139f4c9dad9/src/core/metric/quantized_integer_metric_batch.h#L293-L320)) 

   **`MipsSquaredEuclidean` batch path missing `int8_sum` correction**

   `MipsSquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>::ComputeBatch` uses `ailego::DistanceBatch::InnerProductDistanceBatch<int8_t>` — the same underlying VNNI-capable kernel used by `SquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>`. On IcelAke, this kernel computes the inner product treating `int8` as `uint8` (offset-by-128 trick), which requires the post-computation correction:

   ```cpp
   result -= 128 * int8_sum;
   ```

   `SquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>` correctly applies this correction (line ~222) when `ImplType::GetQueryPreprocessFunc() != nullptr`. `MipsSquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>` does **not** apply any such correction, even though it uses the same kernel. If invoked on IcelAke, the computed distances will be systematically wrong.

   Note: this path is currently unreachable through `BaseDistanceBatchWithScoreUnquantized` (which dispatches `MipsSquaredEuclidean` to `_ComputeBatch`), so there is no immediate runtime regression. However, the struct should either be fixed for consistency or clearly annotated as intentionally incomplete/unused.

2. `src/core/metric/quantized_integer_metric_batch.h`, line 293-320 ([link](https://github.com/alibaba/zvec/blob/81ee36ae30af2ab1daaf9cc3bc274f001ad64c3f/src/core/metric/quantized_integer_metric_batch.h#L293-L320)) 

   **Missing `int8_sum` correction in `MipsSquaredEuclidean` batch path**

   `MipsSquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>::ComputeBatch` calls `ailego::DistanceBatch::InnerProductDistanceBatch<int8_t>` (the same VNNI-capable kernel used by `SquaredEuclidean`), but unlike its sibling, it never applies the `128 * int8_sum` correction.

   When the IcelAke VNNI kernel is active, `InnerProductDistanceBatch` processes the query as unsigned (after preprocessing adds 128 to each element). The raw inner-product result is therefore `Σ(qᵢ + 128) × mᵢ = Σqᵢmᵢ + 128 × int8_sum`. Without the subtraction of `128 * int8_sum`, the `*results` value fed into the final Euclidean formula is wrong, and the computed distance will be incorrect.

   `SquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>::ComputeBatch` applies this correction correctly:
   ```cpp
   if (ImplType::GetQueryPreprocessFunc() != nullptr) {
       int int8_sum = reinterpret_cast<const int *>(m_tail)[4];
       result -= 128 * int8_sum;
   }
   ```

   The same guard + subtraction is missing from `MipsSquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>::ComputeBatch`. While the current dispatch path falls to `_ComputeBatch` (the scalar/matrix path) for `MipsSquaredEuclidean` and this struct is not reachable today, it is a latent bug: the struct's `ComputeBatch` calls the VNNI-capable kernel and would produce silently wrong distances if a dispatch path is ever added.

3. `src/core/metric/quantized_integer_metric_batch.h`, line 293-320 ([link](https://github.com/alibaba/zvec/blob/408dc192b51a054dc6ac38e020075acd462bbcb7/src/core/metric/quantized_integer_metric_batch.h#L293-L320)) 

   **MIPS batch distances wrong on IcelAke: VNNI used without query preprocessing**

   On any hardware with `AVX512_VNNI` (including IcelAke, which is now a valid build target after this PR), `ailego::InnerProductDistanceBatch<int8_t>::ComputeBatch` always dispatches to the VNNI kernel via `_mm512_dpbusd_epi32`. This instruction treats the query register as **unsigned** bytes (`uint8_t`), but `MipsSquaredEuclidean::ComputeBatch` never preprocesses the query (+128 shift), so negative int8 query values (e.g. `-1`) are silently interpreted as their unsigned counterpart (`255`), producing completely wrong inner products.

   Additionally, unlike `SquaredEuclidean` (which was updated in this PR to apply `result -= 128 * int8_sum` when VNNI is active), `MipsSquaredEuclidean::ComputeBatch` has **no** such correction, so the computed distances are doubly incorrect on IcelAke.

   The same issue applies to `MinusInnerProductDistanceBatchWithScoreUnquantized<int8_t>` – `get_query_preprocess_func()` in `quantized_integer_metric.cc` returns `nullptr` for `kInnerProduct`, so no preprocessing is performed; yet on IcelAke the VNNI kernel fires, giving wrong inner products. Oddly, the correction (`result -= 128 * int_sum`) *is* applied inside `MinusInnerProduct::ComputeBatch` whenever `ImplType::GetQueryPreprocessFunc() != nullptr` (i.e. always on VNNI hardware), compounding the error on an already-wrong value.

   Before this PR there was no `ENABLE_ICELAKE` CMake option, so these paths weren't reachable via an explicit IcelAke build. Now they are. Both `kMipsSquaredEuclidean` and `kInnerProduct` INT8 batch distances need either:
   - Their own `QueryPreprocess` exposed through `get_query_preprocess_func()` + the `result -= 128 * int8_sum` correction in `MipsSquaredEuclidean::ComputeBatch`, or
   - A runtime guard that falls back to the non-VNNI path for these metrics.

4. `src/core/metric/quantized_integer_metric_matrix.h`, line 36-59 ([link](https://github.com/alibaba/zvec/blob/408dc192b51a054dc6ac38e020075acd462bbcb7/src/core/metric/quantized_integer_metric_matrix.h#L36-L59)) 

   **`SquaredEuclidean<int8_t>` matrix path silently ignores the new `int8_sum` slot**

   The format now stores 5 metadata units per INT8 vector (`extras[0..3]` as floats + `int8_sum` as `int32_t`), but the matrix `Compute` path only reads four (`ma`, `mb`, `ms`, `ms2` at offsets 0–12). The fifth slot (`int8_sum` at byte offset 16 from the extras base) is silently skipped.

   This is correct in the current code because the matrix path is used exclusively for stored-vector-to-stored-vector distance (during HNSW graph construction) and does not receive a preprocessed query. However, there is no comment explaining this assumption. If `ailego::InnerProductMatrix<int8_t>` is ever updated to also use VNNI for the matrix path (which would require preprocessing), this omission would silently produce wrong distances.

   Consider adding a comment like:
   ```cpp
   // NOTE: int8_sum (at extras offset 16) is intentionally not read here.
   // InnerProductMatrix does not use VNNI query preprocessing, so no
   // "result -= 128 * int8_sum" correction is needed for this path.
   ```

5. `src/core/metric/quantized_integer_metric_matrix.h`, line 36-59 ([link](https://github.com/alibaba/zvec/blob/c1e921e0cd98f711f65e6dd822de41991b34d521/src/core/metric/quantized_integer_metric_matrix.h#L36-L59)) 

   **SquaredEuclidean matrix path missing VNNI int8_sum correction**

   `SquaredEuclidean<int8_t, M, N>::Compute` uses `ailego::InnerProductMatrix` for the raw inner product but never reads or subtracts `128 * int8_sum` from the result. In contrast, the batch path (`SquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>::ComputeBatch`) now conditionally applies:

   ```cpp
   if (ImplType::GetQueryPreprocessFunc() != nullptr) {
       int int8_sum = reinterpret_cast<const int *>(m_tail)[4];
       result -= 128 * int8_sum;
   }
   ```

   If `ailego::InnerProductMatrix` also uses VNNI instructions on IcelAke hardware (shifting query values by +128 to treat them as unsigned), then the bias of `128 * int8_sum` would silently accumulate in all matrix-path distance results, causing incorrect rankings from `distance_matrix()`.

   The new `int8_sum` slot at `(d+16)*M + j*4` is now populated for all INT8 vectors (via the `RecordQuantizer` change), so the data is available to apply the correction. If `ailego::InnerProductMatrix` does NOT use VNNI, a comment explaining this asymmetry would prevent future maintainers from introducing the correction incorrectly.

6. `src/core/metric/quantized_integer_metric_batch.h`, line 287-328 ([link](https://github.com/alibaba/zvec/blob/c1e921e0cd98f711f65e6dd822de41991b34d521/src/core/metric/quantized_integer_metric_batch.h#L287-L328)) 

   **`MipsSquaredEuclidean` batch path lacks int8_sum VNNI correction and `GetQueryPreprocessFunc`**

   `MipsSquaredEuclideanDistanceBatchWithScoreUnquantized<int8_t>` has two gaps compared to the parallel `SquaredEuclidean` specialization that this PR wires up for IcelAke:

   1. **Missing `GetQueryPreprocessFunc`**: The struct defines `QueryPreprocess` but not `GetQueryPreprocessFunc()`. Without the latter, `quantized_integer_metric.cc`'s `get_query_preprocess_func()` cannot register an IcelAke preprocessing function for this metric path, making `QueryPreprocess` unreachable dead code.

   2. **Missing int8_sum correction in `ComputeBatch`**: The loop reads `m_tail[0..3]` (ma, mb, ms, ms2) but never reads `m_tail[4]` (`int8_sum`). If IcelAke preprocessing were ever enabled for MIPS, the uncorrected bias of `128 * int8_sum` from the VNNI inner product would produce silently wrong distances.

   While MIPS + IcelAke is not wired up by this PR, the `QueryPreprocess` method and the `dim - 20` dimension offset both imply it was intended to be. Leaving the struct in an inconsistent half-implemented state is a correctness hazard for future contributors.

7. `src/core/quantizer/record_quantizer.h`, line 78-87 ([link](https://github.com/alibaba/zvec/blob/c1e921e0cd98f711f65e6dd822de41991b34d521/src/core/quantizer/record_quantizer.h#L78-L87)) 

   **INT4 inner-product path still writes `int8_sum` as `int *` alias**

   After the refactor, the INT4 non-euclidean branch still casts through `int *` for its write:

   ```cpp
   reinterpret_cast<int *>(extras)[3] = int8_sum;
   ```

   whereas the new INT8 branch uses `int32_t *`. Using `int *` relies on `sizeof(int) == 4` being universally true, which holds in practice on all relevant targets but is not guaranteed by the C++ standard. Prefer `reinterpret_cast<int32_t *>` for consistency with the INT8 path and explicitness:

</details>

<sub>Last reviewed commit: c1e921e</sub>

### DIFF
--- a/tests/core/metric/quantized_integer_metric_test.cc
+++ b/tests/core/metric/quantized_integer_metric_test.cc
@@ -251,7 +251,7 @@ void TestDistanceMatrixInt8(const std::string &metric_name) {
 
   const size_t batch_size = M;
   const size_t query_size = N;
-  size_t dimension = (std::uniform_int_distribution<size_t>(1, 65))(gen) * 4;
+  size_t dimension = (std::uniform_int_distribution<size_t>(1, 65))(gen)*4;
   auto holder = GetHolder(dimension, batch_size, dist);
   IndexMeta meta(IndexMeta::DT_FP32, dimension);
   meta.set_metric(metric_name, 0, Params());
@@ -261,7 +261,7 @@ void TestDistanceMatrixInt8(const std::string &metric_name) {
   ASSERT_EQ(0u, IndexConverter::TrainAndTransform(converter, holder));
   auto holder2 = converter->result();
   auto &meta2 = converter->meta();
-  ASSERT_EQ(dimension + 16, holder2->dimension());
+  ASSERT_EQ(dimension + 20, holder2->dimension());
   size_t matrix_size = batch_size * holder2->dimension();
   std::vector<int8_t> matrix1(matrix_size);
   std::vector<int8_t> matrix2(matrix_size);
@@ -277,7 +277,7 @@ void TestDistanceMatrixInt8(const std::string &metric_name) {
   auto query_holder = GetHolder(dimension, query_size, dist);
   ASSERT_EQ(0u, IndexConverter::TrainAndTransform(converter, query_holder));
   auto query_holder2 = converter->result();
-  ASSERT_EQ(dimension + 16, query_holder2->dimension());
+  ASSERT_EQ(dimension + 20, query_holder2->dimension());
   size_t query_matrix_size = query_size * query_holder2->dimension();
   std::vector<int8_t> query1(query_matrix_size);
   std::vector<int8_t> query2(query_matrix_size);
@@ -453,7 +453,7 @@ void TestDistanceMatrixInt4(const std::string &metric_name) {
 
   const size_t batch_size = M;
   const size_t query_size = N;
-  size_t dimension = (std::uniform_int_distribution<size_t>(1, 65))(gen) * 8;
+  size_t dimension = (std::uniform_int_distribution<size_t>(1, 65))(gen)*8;
   auto holder = GetHolder(dimension, batch_size, dist);
   IndexMeta meta(IndexMeta::DT_FP32, dimension);
   meta.set_metric(metric_name, 0, Params());


luoxiaojian:
@greptile

richyreachy:
@greptile

luoxiaojian:
@greptile

richyreachy (approved):

PR #226, opened 2026-03-14
fix/fix mips euclidean

richyreachy:
fix mips euclidean which will crash in clang on AMD T7

<h3>Greptile Summary</h3>

This PR refactors the MIPS euclidean distance computation across all data types (fp32, fp16, int4, int8) to fix a crash in Clang on AMD T7. The root cause was inline unary negation and a convoluted dangling-`else` dispatch chain that Clang misoptimised. The fix introduces per-architecture high-level injection wrapper functions (`MipsEucldeanDistanceSphericalInjection*` / `MipsEucldeanDistanceRepeatedQuadraticInjection*`) so each dispatch file calls a single named function and returns early, eliminating the problematic conditional chain. The same pattern is applied to `MinusInnerProduct*` wrappers for the inner-product matrix path.

**Key changes:**
- Adds `MinusInnerProduct{SSE,AVX,AVX512,NEON}` wrappers replacing inline `-InnerProduct*(...)` calls in the `MinusInnerProductMatrix` dispatch.
- Adds `MipsEucldeanDistance{Spherical,RepeatedQuadratic}Injection{SSE,AVX,AVX512,NEON,AVX2}` for fp32, fp16, int4, int8.
- Moves ARM NEON `float` specialisations from `fp32_neon.cc` into `fp32_dispatch.cc` under `#if defined(__ARM_NEON)`.
- Renames `InnerProductAndSquaredNormAVX` → `InnerProductAndSquaredNormAVX2` in the int4 path (fixing a prior naming inconsistency).
- Tightens the SSE guard in int4/int8 dispatch from `__SSE__` to `__SSE4_1__` to match the implementation file.

**Issue found:**
- In `mips_euclidean_distance_matrix_fp16_neon.cc`, the two new NEON injection wrappers are placed *outside* the `#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)` guard, yet they call `InnerProductAndSquaredNormNEON` (Float16 overload) which is compiled only *inside* that guard. On AArch64 targets without `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` this produces a linker error. Both wrappers must be moved inside the `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` block.

- Not safe to merge — the fp16 NEON injection wrappers reference an undefined symbol on AArch64 targets without FP16 vector arithmetic, causing a linker error.
- The x86 refactoring (fp32/fp16/int4/int8 AVX/AVX512/SSE) is well-structured and resolves the reported crash. However, the fp16 ARM path introduces a concrete linker error: the new NEON wrappers are compiled outside `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` while calling a function that is only defined inside it. This will break any AArch64 build that lacks FP16 vector arithmetic support. The missing AVX runtime check in `fp16_dispatch.cc` is a secondary concern and is safe in practice since `__F16C__` implies AVX hardware.
- Primary: `src/ailego/math/mips_euclidean_distance_matrix_fp16_neon.cc` — new wrappers must be moved inside `#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)`. Secondary: `src/ailego/math/mips_euclidean_distance_matrix_fp16_dispatch.cc` — AVX fallback consistency with fp32 dispatch.

| Filename | Overview |
|----------|----------|
| src/ailego/math/mips_euclidean_distance_matrix_fp16_neon.cc | New NEON injection wrappers placed outside the `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` guard but call `InnerProductAndSquaredNormNEON` (Float16) which is only compiled within that guard — linker error on AArch64 without FP16 vector arithmetic. |
| src/ailego/math/mips_euclidean_distance_matrix_fp16_dispatch.cc | Dispatch refactored to call new high-level injection wrappers; AVX512 runtime guard retained, but AVX fallback (lines 58/75) no longer checks `CpuFeatures::static_flags_.AVX`, unlike the fp32 dispatch. |
| src/ailego/math/mips_euclidean_distance_matrix_fp32_dispatch.cc | Clean refactor: x86 dispatch now delegates to typed injection wrappers with proper early-return pattern; ARM NEON overloads moved in from neon.cc with zero-initialised locals; SSE forward declarations correctly use `__SSE__` guard matching the call site. |
| src/ailego/math/mips_euclidean_distance_matrix_fp32_avx.cc | New AVX injection wrappers correctly fall back to SSE for small dimensions (size ≤ 7); SSE forward declaration inside `__SSE__` guard is appropriate as AVX implies SSE. |
| src/ailego/math/mips_euclidean_distance_matrix_int4_dispatch.cc | Guard updated from `__AVX__` to `__AVX2__` for AVX2 forward declarations; SSE guard tightened from `__SSE__` to `__SSE4_1__` to match the sse.cc implementation guard; dispatch correctly uses early-return pattern. |
| src/ailego/math/inner_product_matrix_fp32_dispatch.cc | Dispatch updated to call `MinusInnerProduct*` wrappers instead of inline unary negation; forward declarations correctly guard all new symbols. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["MipsSquaredEuclideanDistanceMatrix::Compute()\nDispatch file"] --> B{Platform}

    B -->|__ARM_NEON| C["MipsEucldeanDistanceSphericalInjectionNEON()\nfp16_neon.cc / fp32_dispatch.cc"]
    C --> D["InnerProductAndSquaredNormNEON()\n⚠️ only when __ARM_FEATURE_FP16_VECTOR_ARITHMETIC\n(fp16 path)"]

    B -->|x86| E{Runtime CPU check}
    E -->|AVX512F available| F["MipsEucldeanDistanceSphericalInjectionAVX512()\nfp*_avx512.cc"]
    F --> G{dim check}
    G -->|> 15| H["InnerProductAndSquaredNormAVX512()"]
    G -->|> 7| I["InnerProductAndSquaredNormAVX()"]
    G -->|≤ 7| J["InnerProductAndSquaredNormSSE()"]

    E -->|AVX available| K["MipsEucldeanDistanceSphericalInjectionAVX()\nfp*_avx.cc"]
    K --> L{dim check}
    L -->|> 7| M["InnerProductAndSquaredNormAVX()"]
    L -->|≤ 7| N["InnerProductAndSquaredNormSSE()"]

    E -->|SSE only| O["MipsEucldeanDistanceSphericalInjectionSSE()\nfp*_sse.cc"]
    O --> P["InnerProductAndSquaredNormSSE()"]

    F --> Q["ComputeSphericalInjection()"]
    K --> Q
    O --> Q
    C --> Q
```

<details open><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `src/ailego/math/inner_product_matrix_fp32_dispatch.cc`, line 47-50 ([link](https://github.com/alibaba/zvec/blob/05e0ac6aa4fd00d3712e132edc0216235027cee8/src/ailego/math/inner_product_matrix_fp32_dispatch.cc#L47-L50)) 

   **Debug/development leftover: commented-out pragma**

   This blank area containing a commented-out `#pragma GCC novector` appears to be a debug artefact that was not cleaned up before submitting. It should be removed.

2. `src/ailego/math/mips_euclidean_distance_matrix_fp32_dispatch.cc`, line 606-607 ([link](https://github.com/alibaba/zvec/blob/05e0ac6aa4fd00d3712e132edc0216235027cee8/src/ailego/math/mips_euclidean_distance_matrix_fp32_dispatch.cc#L606-L607)) 

   **Uninitialized local variables in ARM NEON path**

   `u2` and `v2` are declared uninitialized here, while every other new helper function in this PR (and all the pre-existing ones) initialise them to `0.0f`. If `InnerProductAndSquaredNormNEON` is ever called with `size == 0`, none of the loop/switch paths execute, leaving the out-parameter pointers unwritten — and the garbage values would be forwarded to `ComputeSphericalInjection`, producing undefined behaviour.

3. `src/ailego/math/mips_euclidean_distance_matrix_fp16_neon.cc`, line 333-366 ([link](https://github.com/alibaba/zvec/blob/46449837cba3ea42550cde899c55dc212f03c17b/src/ailego/math/mips_euclidean_distance_matrix_fp16_neon.cc#L333-L366)) 

   **New NEON wrappers outside `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` guard**

   `MipsEucldeanDistanceSphericalInjectionNEON` and `MipsEucldeanDistanceRepeatedQuadraticInjectionNEON` are placed *after* the `#endif  // __ARM_FEATURE_FP16_VECTOR_ARITHMETIC` closing brace (line 331), but both functions call `InnerProductAndSquaredNormNEON(lhs, rhs, size, &u2, &v2)` with `Float16 *` arguments. That overload of `InnerProductAndSquaredNormNEON` is defined *inside* the `#if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)` block (before line 331).

   The dispatch file (`fp16_dispatch.cc`) is compiled when `__ARM_NEON && __aarch64__` is defined, and it directly calls these new NEON wrappers. On an AArch64 target that defines `__ARM_NEON` and `__aarch64__` but **not** `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` (e.g. older Cortex-A57/A72), the wrappers will compile but `InnerProductAndSquaredNormNEON` for `Float16` will not be defined, causing a **linker error** (undefined reference).

   The fix is to move both new functions inside the `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` guard:

   ```cpp
   // Inside #if defined(__ARM_FEATURE_FP16_VECTOR_ARITHMETIC)

   float MipsEucldeanDistanceSphericalInjectionNEON(const Float16 *lhs, ...) { ... }
   float MipsEucldeanDistanceRepeatedQuadraticInjectionNEON(const Float16 *lhs, ...) { ... }

   #endif  // __ARM_FEATURE_FP16_VECTOR_ARITHMETIC
   #endif  // __ARM_NEON && __aarch64__
   ```

</details>

<sub>Last reviewed commit: 4644983</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/src/db/collection.cc
+++ b/src/db/collection.cc
@@ -1722,7 +1722,7 @@ Status CollectionImpl::create() {
   }
   if (ailego::FileHelper::IsExist(path_.c_str())) {
     return Status::InvalidArgument("path validate failed: path[", path_,
-                                   "] is existed");
+                                   "] exists");
   }
 
   // check schema
--- a/src/db/index/column/vector_column/combined_vector_column_indexer.cc
+++ b/src/db/index/column/vector_column/combined_vector_column_indexer.cc
@@ -104,11 +104,22 @@ Result<IndexResults::Ptr> CombinedVectorColumnIndexer::Search(
       need_refine = true;
     }
 
+    const IndexFilter *filter{nullptr};
+    auto per_block_filter =
+        BlockOffsetFilter{query_params.filter, block_offsets_[i]};
+    if (query_params.filter) {
+      if (block_offsets_[i] > 0) {
+        filter = &per_block_filter;
+      } else {
+        filter = query_params.filter;
+      }
+    }
+
     vector_column_params::QueryParams modified_query_params{
         query_params.data_type,
         query_params.dimension,
         query_params.topk,
-        query_params.filter,
+        filter,
         query_params.fetch_vector,
         query_params.query_params,
         query_params.group_by
--- a/src/db/index/column/vector_column/combined_vector_column_indexer.h
+++ b/src/db/index/column/vector_column/combined_vector_column_indexer.h
@@ -15,9 +15,9 @@
 
 #include <memory>
 #include <vector>
+#include "db/index/common/index_filter.h"
 #include "vector_column_indexer.h"
 #include "vector_column_params.h"
-#include "vector_index_results.h"
 
 namespace zvec {
 
@@ -42,8 +42,31 @@ class CombinedVectorColumnIndexer {
   virtual Result<vector_column_params::VectorDataBuffer> Fetch(
       uint32_t segment_doc_id) const;
 
-  // for ut
+
  protected:
+  /**
+   * A filter wrapper that applies an offset to document IDs before
+   * delegating to an inner filter.
+   *
+   * This is used when multiple blocks with different ID offsets are stored.
+   * Each block has its own local ID space, and this filter translates
+   * block-level IDs to segment-level IDs before checking the inner filter.
+   */
+  class BlockOffsetFilter : public IndexFilter {
+   public:
+    BlockOffsetFilter(const IndexFilter *inner_filter, uint64_t offset)
+        : inner_filter_(inner_filter), offset_(offset) {}
+
+    bool is_filtered(uint64_t id) const override {
+      return inner_filter_->is_filtered(id + offset_);
+    }
+
+   private:
+    const IndexFilter *inner_filter_;
+    uint64_t offset_;
+  };
+
+  // for ut
   CombinedVectorColumnIndexer() = default;
 
 
--- a/src/db/index/segment/segment.cc
+++ b/src/db/index/segment/segment.cc
@@ -302,14 +302,14 @@ class SegmentImpl : public Segment,
   void fresh_persist_chunked_array();
 
  private:
-  // scalar forward
+  // scalar forward (uses segment-local doc ID)
   MemForwardStore::Ptr memory_store_;
   std::vector<BaseForwardStore::Ptr> persist_stores_;
 
-  // scalar index
+  // scalar index (uses segment-local doc ID)
   InvertedIndexer::Ptr invert_indexers_;
 
-  // vector index
+  // vector index (uses block-local doc ID, each indexer starts from 0)
   std::unordered_map<std::string, VectorColumnIndexer::Ptr>
       memory_vector_indexers_;
 
@@ -339,7 +339,7 @@ class SegmentImpl : public Segment,
   IDMap::Ptr id_map_;
   DeleteStore::Ptr delete_store_;
 
-  // local_id(index) -> global_doc_id(value)
+  // Maps segment-local doc ID (array index) to global doc ID (stored value)
   std::vector<uint64_t> doc_ids_;
 
   std::array<std::variant<std::vector<int>,
--- a/tests/db/sqlengine/recall_base.h
+++ b/tests/db/sqlengine/recall_base.h
@@ -253,6 +253,7 @@ inline Segment::Ptr RecallTest::create_segment() {
   SegmentOptions options;
   options.read_only_ = false;
   options.enable_mmap_ = true;
+  options.max_buffer_size_ = 256 * 1024;
 
   auto result =
       Segment::CreateAndOpen(GetPath(), *collection_schema_, 0, 0, id_map,
--- a/tests/db/sqlengine/vector_recall_test.cc
+++ b/tests/db/sqlengine/vector_recall_test.cc
@@ -212,4 +212,71 @@ TEST_F(VectorRecallTest, Sparse) {
   }
 }
 
+TEST_F(VectorRecallTest, DeleteFilter) {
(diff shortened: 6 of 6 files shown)


richyreachy:
@greptile

richyreachy:
@greptile

richyreachy:
@greptile

src/ailego/math/inner_product_matrix_fp32_dispatch.cc
```diff
@@ -37,6 +38,10 @@ float InnerProductSSE(const float *lhs, const float *rhs, size_t size);
 float MinusInnerProductSSE(const float *lhs, const float *rhs, size_t size);
 #endif
 
+
+// #pragma GCC novector
```
egolearner: Is this needed?

richyreachy: done

richyreachy:
@greptile

PR #234, opened 2026-03-17
feat(c-api): add nullable setter and per-doc write results

kdy1:
## Summary
- add `zvec_doc_set_field_null` for explicit nullable writes
- add per-document write result APIs for insert/update/upsert/delete
- add allocator-safe free API (`zvec_free_ptr`) and tests

## Tests
- `cmake --build build-capi --target c_api_test -j4`
- `ctest --output-on-failure -R c_api_test`

<h3>Greptile Summary</h3>

This PR extends the C API with four new `_with_results` DML variants (insert/update/upsert/delete) that return per-document `ZVecWriteResult` structs, a `zvec_doc_set_field_null` setter for explicit nullable writes, a generic `zvec_free_ptr` helper, and fetch-side normalization of nullable fields. The overall design and API shape are sound and consistent with the existing C API patterns.

Key concerns:
- **P0 – Crash on OOM in `copy_string`**: The pre-existing `copy_string` helper does not check if `malloc` returns null before calling `strcpy`, causing a null-pointer dereference. This is now called `2×N` times per batch write in `build_write_results`, making it realistically reachable under memory pressure.
- **P1 – Null `pk` in `ZVecWriteResult`**: `copy_string` returns `nullptr` for empty strings, so documents without a PK silently produce a `nullptr` in `results[i].pk`. The struct documentation does not call this out, risking unexpected null dereferences in caller code.
- **P2 – Silent schema-fetch failure during fetch normalization**: If `Schema()` fails while normalizing nullable fields on fetch, the function silently skips normalization and returns data without null markers, directly undermining the feature's correctness guarantee.
- **P2 – `zvec_free_ptr` test doesn't exercise cross-DLL safety**: The test allocates with the caller's `malloc` rather than with a zvec API, so it does not validate the allocator-symmetry the function is designed to provide.

- Not safe to merge as-is — a null pointer dereference in copy_string can crash the process under OOM, and empty PKs silently produce null fields in the result array.
- The P0 crash path in copy_string (no malloc-failure check before strcpy) is now amplified by the new batch loop in build_write_results. The P1 silent-null-pk issue can cause unexpected null dereferences in callers. Both must be fixed before merging.
- src/c_api/c_api.cc — specifically copy_string (line 793) and build_write_results (lines 820-851)

| Filename | Overview |
|----------|----------|
| src/c_api/c_api.cc | Adds nullable setter, per-doc write result helpers, and fetch-side nullable normalization. Critical: `copy_string` crashes on malloc failure (null deref via strcpy), and empty PKs silently produce null entries in result arrays. |
| src/include/zvec/c_api.h | Declares ZVecWriteResult struct and four new _with_results DML APIs plus zvec_free_ptr / zvec_doc_set_field_null. API shape and documentation are clean; the ZVecWriteResult.pk field should note it may be null. |
| tests/c_api/c_api_test.c | Good coverage for null-field API, nullable roundtrip, and DML result APIs. Minor: zvec_free_ptr test uses caller-allocated buffer instead of API-allocated memory, which doesn't validate the intended cross-DLL safety. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Caller as C Caller
    participant API as C API (c_api.cc)
    participant Coll as zvec::Collection

    Note over Caller,Coll: _with_results DML flow (insert/update/upsert)
    Caller->>API: zvec_collection_upsert_with_results(collection, docs, N, &results, &count)
    API->>API: collect_doc_pks(docs, N) → pks[]
    API->>API: convert_zvec_docs_to_internal(docs, N) → internal_docs[]
    API->>Coll: Upsert(internal_docs)
    Coll-->>API: Expected<vector<Status>>
    API->>API: handle_expected_result(result)
    alt operation failed
        API-->>Caller: ZVEC_ERROR_* (results=nullptr, count=0)
    else operation succeeded
        API->>API: build_write_results(statuses, pks, &results, &count)
        Note over API: calloc(N, sizeof(ZVecWriteResult))<br/>loop: copy_string(pk), copy_string(message)
        API-->>Caller: ZVEC_OK, results[N], count=N
        Caller->>API: zvec_write_results_free(results, count)
        API->>API: free_write_results_internal(results, count)
    end

    Note over Caller,Coll: fetch + nullable normalization flow
    Caller->>API: zvec_collection_fetch(collection, pks, N, &docs, &count)
    API->>Coll: Fetch(pk_vector)
    Coll-->>API: Expected<DocPtrMap>
    API->>Coll: Schema()
    Coll-->>API: Expected<CollectionSchema>
    alt schema available
        API->>API: normalize_nullable_fields_for_fetch(schema, doc_map)
        Note over API: For each nullable field absent from a doc,<br/>call doc->set_null(field_name)
    else schema fetch failed
        API->>API: (silently skip normalization)
    end
    API->>API: convert_fetched_document_results(doc_map, &docs, &count)
    API-->>Caller: ZVEC_OK, docs[N]
```

1. `src/c_api/c_api.cc`, line 793-799 ([link](https://github.com/alibaba/zvec/blob/f137932458750ff831d14b6d704139bef61a32c1/src/c_api/c_api.cc#L793-L799)) 

   <a href="#">P0</a> **Null dereference crash on malloc failure in `copy_string`**

   When `malloc` returns `nullptr` (OOM), `strcpy(copy, str.c_str())` is called with a null destination pointer, causing a **segfault/undefined behavior** instead of returning an error. This is a pre-existing bug, but `build_write_results` now calls `copy_string` in a tight loop — **2×N times** for a batch of N documents — making this path reachable under memory pressure in production:

   ```cpp
   char *copy = static_cast<char *>(malloc(str.length() + 1));
   strcpy(copy, str.c_str());  // UB if malloc returned nullptr
   ```

   Fix by checking before dereferencing:

   

   Additionally, `build_write_results` should then detect a null `pk`/`message` from a non-empty input and clean up the partially-built array before returning `ZVEC_ERROR_INTERNAL_ERROR`.

</details>

<sub>Last reviewed commit: f137932</sub>

> Greptile also left **3 inline comments** on this PR.

### DIFF
--- a/src/c_api/c_api.cc
+++ b/src/c_api/c_api.cc
@@ -798,6 +798,73 @@ static char *copy_string(const std::string &str) {
   return copy;
 }
 
+// Helper function: free write results returned by detailed DML APIs.
+static void free_write_results_internal(ZVecWriteResult *results,
+                                        size_t result_count) {
+  if (!results) {
+    return;
+  }
+  for (size_t i = 0; i < result_count; ++i) {
+    if (results[i].pk) {
+      free((void *)results[i].pk);
+      results[i].pk = nullptr;
+    }
+    if (results[i].message) {
+      free((void *)results[i].message);
+      results[i].message = nullptr;
+    }
+  }
+  free(results);
+}
+
+// Helper function: convert per-doc statuses to C API write result array.
+static ZVecErrorCode build_write_results(
+    const std::vector<zvec::Status> &statuses,
+    const std::vector<std::string> &pks, ZVecWriteResult **results,
+    size_t *result_count) {
+  if (!results || !result_count) {
+    return ZVEC_ERROR_INVALID_ARGUMENT;
+  }
+
+  *result_count = statuses.size();
+  if (*result_count == 0) {
+    *results = nullptr;
+    return ZVEC_OK;
+  }
+
+  *results = static_cast<ZVecWriteResult *>(
+      calloc(*result_count, sizeof(ZVecWriteResult)));
+  if (!*results) {
+    set_last_error("Failed to allocate memory for write results");
+    return ZVEC_ERROR_INTERNAL_ERROR;
+  }
+
+  for (size_t i = 0; i < *result_count; ++i) {
+    const std::string pk = i < pks.size() ? pks[i] : std::string();
+    const std::string message = statuses[i].message();
+    (*results)[i].pk = copy_string(pk);
+    (*results)[i].message = copy_string(message);
+    (*results)[i].code = status_to_error_code(statuses[i]);
+  }
+
+  return ZVEC_OK;
+}
+
+static std::vector<std::string> collect_doc_pks(const ZVecDoc **docs,
+                                                size_t doc_count) {
+  std::vector<std::string> pks;
+  pks.reserve(doc_count);
+  for (size_t i = 0; i < doc_count; ++i) {
+    if (!docs[i]) {
+      pks.emplace_back("");
+      continue;
+    }
+    auto doc_ptr = reinterpret_cast<const std::shared_ptr<zvec::Doc> *>(docs[i]);
+    pks.emplace_back((*doc_ptr)->pk_ref());
+  }
+  return pks;
+}
+
 static zvec::DataType convert_data_type(ZVecDataType zvec_type) {
   if (zvec_type < ZVEC_DATA_TYPE_UNDEFINED ||
       zvec_type > ZVEC_DATA_TYPE_ARRAY_DOUBLE) {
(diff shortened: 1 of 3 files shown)


src/c_api/c_api.cc
```diff
@@ -5730,6 +5981,10 @@ default: {
           return ZVEC_ERROR_INTERNAL_ERROR;
         }
 
-        const auto &doc_map = result.value();
+        auto doc_map = result.value();
+        auto schema_result = (*coll_ptr)->Schema();
+        if (schema_result.has_value()) {
+          normalize_nullable_fields_for_fetch(schema_result.value(), doc_map);
```
chinaux, line 5987: Currently, the C++ implementation omits the key from the document if the field is null. However, if you require the document to include all fields regardless, we can certainly accommodate that.

PR #243, opened 2026-03-18
feat: Introduce turbo.

luoxiaojian:
<h3>Greptile Summary</h3>

This PR introduces a new `zvec_turbo` subsystem that provides runtime-dispatched, AVX512-VNNI–accelerated distance computation for INT8-quantized vectors. When the CPU supports AVX512-VNNI, the cosine and squared-Euclidean distance functions (single-vector, batch, and query-preprocessing paths) are transparently replaced with optimized implementations; otherwise the existing generic fallback is used unchanged. The build system correctly applies per-file `-march=avx512vnni` flags only to the relevant source files and skips x86-specific optimizations on Android/ARM64.

Key changes:
- New `src/turbo/` module with AVX512-VNNI inner product kernels (`common.h`), cosine, and squared-Euclidean distance implementations
- Public API exposed via `src/include/zvec/turbo/turbo.h` with three dispatch functions
- `quantized_integer_metric.cc` wired to probe turbo first, then fall back to the generic path
- All relevant CMake targets updated to link against `zvec_turbo`

Issues found:
- `cosine_int8_batch_distance` uses a **`int i` loop variable** compared against `size_t n`, producing a signed/unsigned mismatch (see inline comment). The companion `squared_euclidean_int8_batch_distance` correctly uses `size_t i`.
- `cosine_int8_distance` and `cosine_int8_batch_distance` compute `original_dim` via an **implicit `size_t → int` narrowing** (`dim - 24`) without an explicit cast, while `cosine_int8_query_preprocess` uses `static_cast<int>(dim) - 24`; the inconsistency introduces technically implementation-defined behavior for tiny `dim` values.
- The **cosine header comment** says "24-byte metadata tail (3 floats)" — 3 floats is only 12 bytes. The `.cc` layout block also omits the 6th 4-byte field, making the actual 24-byte tail appear to be 20 bytes.
- `FMA_INT8_GENERAL` macro in `common.h` lacks a `do { } while(0)` wrapper, which is fragile for future callers outside a switch-case context.

- The fallback path is always preserved, so existing functionality is not broken, but the signed/unsigned loop issue and inconsistent casting in the cosine implementation should be addressed before merge.
- The dispatch logic and squared-Euclidean implementation are solid, but the cosine batch distance function has a signed/unsigned loop variable mismatch and inconsistent `original_dim` computation that differ from the squared-Euclidean counterpart without explanation. The header comment is also internally contradictory. These lower confidence in the cosine-specific code paths.
- `src/turbo/avx512_vnni/record_quantized_int8/cosine.cc` and `cosine.h` need the most attention.

| Filename | Overview |
|----------|----------|
| src/turbo/avx512_vnni/record_quantized_int8/common.h | New shared AVX512-VNNI inner product kernel header; contains the non-VNNI (AVX2) signed int8 dot product and the VNNI `dpbusd`-based batched kernel; minor: macro hygiene and the `abs(-128)` saturation edge case are worth noting. |
| src/turbo/avx512_vnni/record_quantized_int8/cosine.cc | AVX512-VNNI cosine distance implementation; has a signed/unsigned mismatch (`int i` vs `size_t n`) in the batch loop and uses implicit `size_t→int` narrowing for `original_dim` without an explicit cast, unlike the query-preprocess function in the same file. |
| src/turbo/avx512_vnni/record_quantized_int8/cosine.h | Public declarations for AVX512-VNNI cosine distance functions; header comment incorrectly states "24-byte metadata tail (3 floats)" — 3 floats is only 12 bytes, not 24. |
| src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.cc | AVX512-VNNI squared Euclidean distance implementation; uses correct `size_t i` loop variable and `static_cast<int>(dim) - 20` for query preprocessing; logic and tail layout are consistent with existing batch infrastructure. |
| src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.h | Public declarations for AVX512-VNNI squared Euclidean distance functions; minor: header comment omits the `int8_sum` field from the tail description, but code is correct. |
| src/turbo/turbo.cc | Runtime CPU-feature dispatcher that returns the appropriate AVX512-VNNI function pointers when `CpuFeatures::static_flags_.AVX512_VNNI` is set; clean and consistent across all three dispatch functions. |
| src/include/zvec/turbo/turbo.h | New public API header for the turbo subsystem; defines `DistanceFunc`, `BatchDistanceFunc`, `QueryPreprocessFunc` type aliases and the three dispatch functions; clean and well-structured. |
| src/core/metric/quantized_integer_metric.cc | Integration point: inserts turbo accelerated function lookups ahead of the generic fallback for INT8 squared-Euclidean, INT8 cosine (matrix, batch, and query-preprocess paths); pattern is uniform and falls back correctly when turbo returns `nullptr`. |
| src/turbo/CMakeLists.txt | New build target `zvec_turbo`; correctly applies per-file AVX512-VNNI compile flags only to `avx512_vnni/*.cc` sources and skips x86-specific flags on Android/ARM64 targets. |
| src/core/CMakeLists.txt | Adds `zvec_turbo` as a library dependency for `zvec_core`; pre-existing missing newline at end of file is not fixed by this PR. |
| src/CMakeLists.txt | Adds `cc_directory(turbo)` before `cc_directory(core)` so that `zvec_turbo` is built before `zvec_core` which depends on it; ordering is correct. |
| src/core/metric/CMakeLists.txt | Adds `zvec_turbo` dependency to `core_metric`; straightforward change. |
| examples/c++/CMakeLists.txt | Adds `zvec_turbo` to example `zvec_core_deps`; removes the placeholder comment. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["QuantizedIntegerMetric\n(quantized_integer_metric.cc)"] --> B["turbo::get_distance_func /\nget_batch_distance_func /\nget_query_preprocess_func"]
    B --> C{CpuFeatures::\nAVX512_VNNI?}
    C -- Yes --> D{MetricType?}
    D -- kSquaredEuclidean + kInt8 --> E["avx512_vnni::\nsquared_euclidean_int8_distance\nsquared_euclidean_int8_batch_distance\nsquared_euclidean_int8_query_preprocess"]
    D -- kCosine + kInt8 --> F["avx512_vnni::\ncosine_int8_distance\ncosine_int8_batch_distance\ncosine_int8_query_preprocess"]
    E --> G["ip_int8_avx512_vnni\n(AVX2 sign-trick, single-vector)"]
    E --> H["ip_int8_batch_avx512_vnni\n(VNNI dpbusd, batched)"]
    F --> G
    F --> H
    C -- No --> I["nullptr → fallback to\nDistanceMatrixCompute /\nBaseDistanceBatchWithScore..."]
```

<details open><summary><h3>Comments Outside Diff (11)</h3></summary>

1. `src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.cc`, line 899 ([link](https://github.com/alibaba/zvec/blob/432d541ef1da7103642ee7cb1906ba1b9181dceb/src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.cc#L899)) 

   <a href="#">P1</a> **Integer arithmetic for bias correction**

   The bias correction `result -= 128 * int8_sum` is computed using integer arithmetic (two `int` operands), whereas the corresponding correction in `cosine.cc` line 706 correctly uses `result -= 128.0f * static_cast<float>(int8_sum)`.

   While integer overflow is unlikely in practice (for dimensions ≤ 2048, the product fits easily in `int32`), the inconsistency with the cosine path is confusing and the implicit integer-to-float conversion is subtle. Using float arithmetic here is safer and consistent with the cosine implementation.

2. `src/turbo/avx512_vnni/record_quantized_int8/cosine.cc`, line 619-636 ([link](https://github.com/alibaba/zvec/blob/432d541ef1da7103642ee7cb1906ba1b9181dceb/src/turbo/avx512_vnni/record_quantized_int8/cosine.cc#L619-L636)) 

   <a href="#">P1</a> **Tail layout comment does not match the 24-byte tail size**

   The comment at the top of the file enumerates the tail fields as:

   ```
   [ float scale_a       ]   4 bytes
   [ float bias_a        ]   4 bytes
   [ float sum_a         ]   4 bytes
   [ float square_sum_a  ]   4 bytes
   [ int  int8_sum       ]   4 bytes
   ```

   That totals **20 bytes**, yet both `cosine_int8_distance` and `cosine_int8_batch_distance` derive `original_dim` as `dim - 24`, implying a **24-byte** tail. The mismatch of 4 bytes means the comment is missing at least one field. This discrepancy is reinforced by `reinterpret_cast<const int *>(m_tail)[4]` correctly accessing `int8_sum` at byte offset 16 from the tail base (consistent with 4 floats preceding it), but then the tail base itself is 4 bytes earlier than a 20-byte-tail layout would place it.

   As written, the comment will mislead anyone trying to understand or audit the encoding format, and it makes it impossible to verify the `dim - 24` constant without inspecting the encoder. Please document the missing field so the tail size and the constant agree.

3. `src/include/zvec/turbo/turbo.h`, line 194-197 ([link](https://github.com/alibaba/zvec/blob/432d541ef1da7103642ee7cb1906ba1b9181dceb/src/include/zvec/turbo/turbo.h#L194-L197)) 

   <a href="#">P1</a> **`size_t dim` in the function type aliases mismatches `int dim` in all implementations**

   `DistanceFunc` and `BatchDistanceFunc` declare the dimension parameter as `size_t`:

   ```cpp
   using DistanceFunc = std::function<void(const void *m, const void *q, size_t dim, float *out)>;
   using BatchDistanceFunc = std::function<void(
       const void **m, const void *q, size_t num, size_t dim, float *out)>;
   ```

   Yet every concrete implementation (`cosine_int8_distance`, `squared_euclidean_int8_distance`, `cosine_int8_batch_distance`, `squared_euclidean_int8_batch_distance`) declares the corresponding parameter as `int dim` and `int n`. When these function pointers are wrapped into `std::function`, the call site will silently narrow the `size_t` to `int`, which is undefined behaviour for values ≥ 2³¹ on 64-bit platforms (extremely unlikely in practice, but still a latent issue). Additionally, `const void **` vs the implementations' `const void *const *` drops const-correctness on the pointer-to-pointer.

   Consider changing the function-type aliases to match the actual signatures, or update all implementations to accept `size_t`.

4. `src/turbo/avx512_vnni/record_quantized_int8/cosine.cc`, line 707 ([link](https://github.com/alibaba/zvec/blob/c4f6baa192a7ab1cda32757c3b30ec55d865bd93/src/turbo/avx512_vnni/record_quantized_int8/cosine.cc#L707)) 

   **Integer overflow in `128 * int8_sum`**

   `int8_sum` is a 32-bit `int` holding the sum of all `original_dim` raw `int8_t` values. For `original_dim ≥ 131 073` with maximally-magnitude elements (e.g., all `-128`), the product `128 * int8_sum` will overflow `INT_MAX = 2 147 483 647`:

   - max `|int8_sum|` = `original_dim × 128`
   - overflow threshold: `128 × 131 072 × 128 = 2 147 483 648 > INT_MAX`

   The same issue also appears at `squared_euclidean.cc` line 900.

   The fix is to widen before multiplying by casting to `float` first:

5. `src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.cc`, line 900 ([link](https://github.com/alibaba/zvec/blob/c4f6baa192a7ab1cda32757c3b30ec55d865bd93/src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.cc#L900)) 

   **Integer overflow in `128 * int8_sum`**

   Same integer overflow risk as in `cosine.cc:707`. `128 * int8_sum` overflows `int` when `original_dim ≥ 131 073` with extreme values. The identical fix applies:

6. `src/turbo/avx512_vnni/record_quantized_int8/cosine.h`, line 767-769 ([link](https://github.com/alibaba/zvec/blob/c4f6baa192a7ab1cda32757c3b30ec55d865bd93/src/turbo/avx512_vnni/record_quantized_int8/cosine.h#L767-L769)) 

   **Self-contradictory tail-size comment**

   The comment claims a "24-byte metadata tail (3 floats: scale_a, bias_a, sum_a)", but 3 `float`s are only 12 bytes — not 24. The `cosine.cc` implementation reveals the actual layout is 4 floats (`ma`, `mb`, `ms`, `ms2`) + 1 `int` (`int8_sum`) accessed at byte offset 16, totalling 20 bytes of live data inside a 24-byte tail (4 bytes of padding). The description in the header should be updated to match reality:

   

   Compare with `squared_euclidean.h` which correctly states "20-byte metadata tail (4 floats)".

7. `src/core/CMakeLists.txt`, line 17 ([link](https://github.com/alibaba/zvec/blob/c4f6baa192a7ab1cda32757c3b30ec55d865bd93/src/core/CMakeLists.txt#L17)) 

   **Missing newline at end of file**

   The diff shows `\ No newline at end of file` for `src/core/CMakeLists.txt`. POSIX requires text files to end with a newline, and many tools (e.g., `cat`, `git diff`) behave unexpectedly or emit warnings without it. Please add a trailing newline after the closing parenthesis of the `cc_library` block.

8. `src/turbo/avx512_vnni/record_quantized_int8/cosine.cc`, line 694 ([link](https://github.com/alibaba/zvec/blob/cb3d54a6c86c40e14faa36add49933f72a2bf14c/src/turbo/avx512_vnni/record_quantized_int8/cosine.cc#L694)) 

   **Signed loop variable vs. unsigned `n`**

   `i` is declared as `int` while `n` is `size_t` (unsigned). The comparison `int i < size_t n` triggers an implicit signed-to-unsigned promotion of `i`. While this won't cause incorrect behavior in practice (since `i` is always non-negative), it differs from the companion function `squared_euclidean_int8_batch_distance` which correctly uses `size_t i`, and most compilers will emit a signed/unsigned comparison warning here.

9. `src/turbo/avx512_vnni/record_quantized_int8/cosine.h`, line 767-768 ([link](https://github.com/alibaba/zvec/blob/cb3d54a6c86c40e14faa36add49933f72a2bf14c/src/turbo/avx512_vnni/record_quantized_int8/cosine.h#L767-L768)) 

   **Inconsistent tail-size description in header comment**

   The comment says "24-byte metadata tail (3 floats: scale_a, bias_a, sum_a)" — but 3 floats is only 12 bytes, not 24. The actual encoding (matching the pre-existing `CosineMinusInnerProductDistanceBatchWithScoreUnquantized<int8_t>` in `quantized_integer_metric_batch.h`) uses a 24-byte tail with 6 four-byte fields. The corresponding `cosine.cc` layout comment also omits the 6th field (bytes 20–23), making it appear to describe only 20 bytes.

   Suggest updating both the header and the `.cc` block comment to accurately enumerate all 6 fields (e.g. 4 floats + `int8_sum` + 4 bytes of padding/reserved), so that future readers can understand why the subtraction uses 24 rather than 20.

10. `src/turbo/avx512_vnni/record_quantized_int8/cosine.cc`, line 636-637 ([link](https://github.com/alibaba/zvec/blob/cb3d54a6c86c40e14faa36add49933f72a2bf14c/src/turbo/avx512_vnni/record_quantized_int8/cosine.cc#L636-L637)) 

    **Inconsistent `original_dim` computation style across functions**

    `cosine_int8_distance` and `cosine_int8_batch_distance` compute `original_dim` as an implicit `size_t → int` conversion:

    ```cpp
    const int original_dim = dim - 24;  // dim is size_t; subtraction is unsigned
    ```

    whereas `cosine_int8_query_preprocess` (and both `squared_euclidean` preprocessing helpers) use an explicit cast:

    ```cpp
    const int original_dim = static_cast<int>(dim) - 24;
    ```

    Without the explicit cast, `dim - 24` is evaluated as a `size_t` subtraction first (wrapping to a huge value when `dim < 24`), then that large unsigned value is implicitly narrowed to `int` — which is implementation-defined behavior in C++. The guarded `if (original_dim <= 0)` check then relies on the two's-complement wrap to a negative `int`, which works in practice but is technically UB.

    Consistently using `static_cast<int>(dim) - 24` avoids the UB and makes the intent clear:

11. `src/turbo/avx512_vnni/record_quantized_int8/common.h`, line 316 ([link](https://github.com/alibaba/zvec/blob/cb3d54a6c86c40e14faa36add49933f72a2bf14c/src/turbo/avx512_vnni/record_quantized_int8/common.h#L316)) 

    **Header-local macro should be hygienically scoped**

    `FMA_INT8_GENERAL` is `#define`d and then `#undef`d later in the same header. Because this header is included from multiple `.cc` TUs, the define/undef cycle is safe as written, but defining a non-`do { } while(0)`-wrapped statement macro in a header is fragile. An accidental include order change or a future callee that calls it outside a switch context (where a trailing `;` creates a dangling empty statement) could introduce subtle bugs.

    Consider either converting it to a `static inline` helper or adding a `do { } while(0)` wrapper:

    ```cpp
    #define FMA_INT8_GENERAL(m, q, sum) do { (sum) += static_cast<float>((m) * (q)); } while(0)
    ```

</details>

<sub>Last reviewed commit: ["fix"](https://github.com/alibaba/zvec/commit/cb3d54a6c86c40e14faa36add49933f72a2bf14c)</sub>

### DIFF
--- a/src/turbo/avx512_vnni/record_quantized_int8/cosine.cc
+++ b/src/turbo/avx512_vnni/record_quantized_int8/cosine.cc
@@ -110,7 +110,7 @@ void cosine_int8_batch_distance(const void *const *vectors, const void *query,
     // int8_sum is stored as the 5th int-sized field after the 4 floats.
     int int8_sum = reinterpret_cast<const int *>(m_tail)[4];
     float &result = distances[i];
-    result -= static_cast<float>(128 * int8_sum);
+    result -= 128.0f * static_cast<float>(int8_sum);
 
     // Dequantize and compute cosine distance:
     //   cosine_dist = -(ma * qa * ip + mb * qa * qs + qb * ma * ms
--- a/src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.cc
+++ b/src/turbo/avx512_vnni/record_quantized_int8/squared_euclidean.cc
@@ -108,7 +108,7 @@ void squared_euclidean_int8_batch_distance(const void *const *vectors,
     // int8_sum is stored as the 5th int-sized field after the 4 floats.
     int int8_sum = reinterpret_cast<const int *>(m_tail)[4];
     float &result = distances[i];
-    result -= static_cast<float>(128 * int8_sum);
+    result -= 128.0f * static_cast<float>(int8_sum);
     result = ma * ma * ms2 + sum2 - 2 * ma * qa * result +
              (mb - qb) * (mb - qb) * original_dim +
              2 * (mb - qb) * (ms * ma - sum);


luoxiaojian:
@greptile

luoxiaojian:
@greptile

richyreachy (approved):

PR #246, opened 2026-03-20
fix: recovery from crash during optimization

zhourrr:
<h3>Greptile Summary</h3>

This PR introduces crash-recovery support for the `Optimize()` path: when a previous run was interrupted mid-optimization, stale index files left on disk are now detected and removed before re-creation in `SegmentImpl::create_vector_index`. It also ships a new `collection_optimizer` helper binary and a full end-to-end `OptimizeRecoveryTest`, alongside a batch of spelling fixes (`CreateComapctTask`, `new_segmnet_meta`, `faield`, `create`, `indice`) and a renamed `write_recovery_test` collection to avoid conflicts.

- **Core recovery logic** (`segment.cc`): Four new `FileExists` / `RemoveFile` guards ensure stale index files from a crashed optimization are cleaned before being overwritten. However, `FileHelper::RemoveFile` returns a `bool` and its return value is silently discarded at all four new sites — unlike the existing pattern at line 3423 where failure is propagated as a `Status::InternalError`. If removal fails the subsequent indexer creation may itself fail or silently operate on a corrupted file.
- **Typo / naming fixes**: `CreateComapctTask` → `CreateCompactTask` (factory method + all call sites), `new_segmnet_meta` → `new_segment_meta` (virtual method signatures), `create` → `create_if_missing` (storage `open` interface), `indice` → `index` (fetch parameter). All are mechanical and low-risk.
- **`doc.cc` tolerance widening**: Float tolerance increased 100× and double tolerance 1000× in the production `Doc::ValueEqual` comparator — a concern that was raised in a previous review thread and remains without an explanatory comment.
- **New test** (`optimize_recovery_test.cc`): Exercises the full crash-and-recover cycle; the per-document verification loop calls `Fetch` individually for each of up to 75,000 documents which may make the test slow on CI.

- Safe to merge after addressing the unchecked `RemoveFile` return values in the crash-recovery paths of `segment.cc`.
- The majority of changes are low-risk spelling fixes and test improvements. The one P1 concern is that the four new `FileHelper::RemoveFile` calls in `create_vector_index` ignore the boolean return value, which is inconsistent with the existing error-handling pattern in the same file and could leave the system in a worse state than before recovery if a removal fails silently. The previously-flagged 100×/1000× tolerance widening in `doc.cc` also remains without justification, which reduces confidence in the integrity checks the new test is meant to exercise.
- `src/db/index/segment/segment.cc` (four unchecked `RemoveFile` returns at lines 1639, 1683, 1722, 1801) and `src/db/index/common/doc.cc` (undocumented tolerance widening).

| Filename | Overview |
|----------|----------|
| src/db/index/segment/segment.cc | Adds crash-recovery guards to delete stale index files before re-creation; the return value of `RemoveFile` is not checked in any of the four new call sites, unlike the existing pattern elsewhere in the file. |
| src/db/index/common/doc.cc | Widens float equality tolerance 100× (1e-6→1e-4) and double tolerance 1000× (1e-9→1e-6) in the production `Doc::ValueEqual`; rationale for the new bounds is undocumented. |
| tests/db/crash_recovery/collection_optimizer.cc | New binary that opens an existing collection and runs `Optimize()`; used as the crash-injectable process in `optimize_recovery_test.cc`. |
| tests/db/crash_recovery/optimize_recovery_test.cc | New end-to-end test that inserts 50 K docs, kills the optimizer mid-run, then verifies all 75 K docs individually via single-element `Fetch` calls; the per-document fetch loop is functional but potentially slow. |
| tests/db/crash_recovery/write_recovery_test.cc | Renames collection/dir from `crash_test` to `write_recovery_test`, adds a `256*1024` segment limit to `CollectionOptions`, and improves `ASSERT_TRUE` error messages with `.error().message()`. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Test as optimize_recovery_test
    participant Optimizer as collection_optimizer (child process)
    participant Segment as SegmentImpl::create_vector_index
    participant FS as Filesystem

    Test->>Optimizer: fork() + execvp("--path", dir)
    Optimizer->>Segment: Optimize() → create_vector_index()
    loop For each index block
        Segment->>FS: FileHelper::FileExists(index_file_path)
        alt File exists (crash residue)
            Segment->>FS: FileHelper::RemoveFile(index_file_path)
            Note over Segment,FS: ⚠️ return value not checked
        end
        Segment->>FS: merge_vector_indexer(index_file_path, ...)
    end
    Test->>Optimizer: sleep(4s) → kill(SIGKILL)
    Note over Optimizer: Process terminated mid-optimization
    Test->>Test: Collection::Open(dir) — reopen after crash
    Test->>Test: Verify all docs via Fetch()
    Test->>Optimizer: fork() + execvp (RunOptimizer, no crash)
    Optimizer->>Segment: Optimize() → create_vector_index()
    loop Stale files from killed run
        Segment->>FS: FileExists → RemoveFile → merge_vector_indexer
    end
    Optimizer-->>Test: exit(0)
    Test->>Test: Final verification of all 75K docs
```

1. `tests/db/crash_recovery/write_recovery_test.cc`, line 129 ([link](https://github.com/alibaba/zvec/blob/6e3f28f2010400ec21faa713e775ccb87cc7535d/tests/db/crash_recovery/write_recovery_test.cc#L129)) 

   <a href="#">P1</a> **`SetUp`/`TearDown` reference stale path after rename**

   `dir_path_` was renamed from `"crash_test_db"` to `"write_recovery_test_db"` in this PR, but both `SetUp` (line 129) and `TearDown` (line 134) still attempt to delete the old `"./crash_test_db"` path. As a result:
   - The actual test directory `write_recovery_test_db` is **never cleaned up** between test runs, causing test pollution and potential false-positives on repeated runs.
   - The delete of `crash_test_db` is a no-op since that directory no longer exists.

2. `tests/db/crash_recovery/write_recovery_test.cc`, line 134 ([link](https://github.com/alibaba/zvec/blob/6e3f28f2010400ec21faa713e775ccb87cc7535d/tests/db/crash_recovery/write_recovery_test.cc#L134)) 

   <a href="#">P1</a> **Stale path in `TearDown` after rename**

   Same rename miss as in `SetUp`. The `TearDown` must also reference the updated directory name.

3. `src/db/index/segment/segment.cc`, line 1715-1718 ([link](https://github.com/alibaba/zvec/blob/61f3401168e20164f1d49c99248cb8f3971ccda2/src/db/index/segment/segment.cc#L1715-L1718)) 

   <a href="#">P1</a> **Missing crash residue cleanup for quantized index paths**

   The PR adds `FileExists` + `RemoveFile` guards before building vector index files at lines 1634 and 1678 (for `MakeVectorIndexPath`), but the corresponding quantized index files built via `MakeQuantizeVectorIndexPath` — at this location (non-RabitQ, line ~1715) and the RabitQ path (~line 1787) — receive no such guard.

   The same crash scenario applies: if a crash occurs after the flat index block ID is committed to `segment_meta_` but during the quantized index build, `block_id_allocator_` will be reset to `max_block_id + 1` from the persisted meta on recovery, producing the same `quant_block_id` and therefore the same file path. The orphaned (and possibly corrupted) quantized index file will then be silently re-opened by `merge_vector_indexer` rather than being detected and cleaned up.

   ```cpp
         std::string index_file_path = FileHelper::MakeQuantizeVectorIndexPath(
             path_, column, segment_meta_->id(), quant_block_id);
   +     if (FileHelper::FileExists(index_file_path)) {
   +       LOG_WARN(
   +           "Quantize index file[%s] already exists (possible crash residue); "
   +           "cleaning and overwriting.",
   +           index_file_path.c_str());
   +       FileHelper::RemoveFile(index_file_path);
   +     }
         auto vector_indexer = merge_vector_indexer(
             index_file_path, column, *field_with_new_index_params, concurrency);
   ```

   The same fix should be applied to the RabitQ quantized index path at line ~1787.

4. `src/db/index/common/doc.cc`, line 1153-1166 ([link](https://github.com/alibaba/zvec/blob/61f3401168e20164f1d49c99248cb8f3971ccda2/src/db/index/common/doc.cc#L1153-L1166)) 

   <a href="#">P2</a> **Loose float tolerance may hide data corruption**

   The tolerance for `float` vectors is widened 100× (from `1e-6f` to `1e-4f`) and for `double` vectors 1000× (from `1e-9` to `1e-6`). `Doc::operator==` — and by extension the crash-recovery integration tests — now accepts vectors whose components differ by up to ~0.0001 as "equal".

   While some precision loss during compaction/reconstruction is expected (e.g., quantisation round-trips, format conversions), a 100× relaxation is very coarse for `float32`. Any data corruption that introduces per-element errors below `1e-4f` will be silently accepted by the equality check, defeating the purpose of the data-integrity assertions in the new test.

   Consider tightening the tolerance back towards the original values, or — if a specific lossy path is known to cause up to this level of error — adding a comment explaining the maximum expected error and why it is acceptable, so future readers do not loosen it further.

5. `src/db/index/segment/segment.cc`, line 1715-1718 ([link](https://github.com/alibaba/zvec/blob/61f3401168e20164f1d49c99248cb8f3971ccda2/src/db/index/segment/segment.cc#L1715-L1718)) 

   <a href="#">P1</a> **Missing crash residue check for quantize index file**

   The PR adds `FileHelper::FileExists` / `FileHelper::RemoveFile` guards before creating the flat vector index at lines 1634 and 1678, but the analogous guard is absent for `MakeQuantizeVectorIndexPath` in the non-RABITQ quantize path here.

   If the optimizer crashes **after** `quant_block_id` has been allocated and the flat index file has been fully written, but **during** writing of the quantize index file, the quant file will be a partial crash residue. On the next recovery run, `allocate_block_id()` will produce the same `quant_block_id` (block counter recovered from persistent state), constructing the same path — but `merge_vector_indexer` is called against the pre-existing partial file with no cleanup, potentially causing corruption or an incorrect index.

   The flat index path at line 1678 is protected; the quantize path should be too:

   ```cpp
         std::string index_file_path = FileHelper::MakeQuantizeVectorIndexPath(
             path_, column, segment_meta_->id(), quant_block_id);
   +     if (FileHelper::FileExists(index_file_path)) {
   +       LOG_WARN(
   +           "Quantize index file[%s] already exists (possible crash residue); cleaning "
   +           "and overwriting.",
   +           index_file_path.c_str());
   +       FileHelper::RemoveFile(index_file_path);
   +     }
         auto vector_indexer = merge_vector_indexer(
             index_file_path, column, *field_with_new_index_params, concurrency);
   ```

6. `src/db/index/segment/segment.cc`, line 1715-1718 ([link](https://github.com/alibaba/zvec/blob/12480427aa891d1388adddd4ff93d48ea43f29b5/src/db/index/segment/segment.cc#L1715-L1718)) 

   <a href="#">P1</a> **Crash-residue check missing for quantized index paths**

   The PR adds a `FileHelper::FileExists` + `FileHelper::RemoveFile` guard before building index files in the `QuantizeType::UNDEFINED` branch (lines ~1634–1639) and the flat-vector sub-branch of the quantized case (lines ~1678–1683). However, the same guard is **not** applied here when building the quantized index file via `MakeQuantizeVectorIndexPath`.

   If optimization crashes after the flat/raw vector index is successfully written but before this quantized index file is committed, the next recovery run will reallocate the same `quant_block_id`, construct the same file path, and call `merge_vector_indexer` on an already-existing (potentially partial) file. The resulting behavior—silently overwriting, failing, or producing a corrupt index—is undefined.

   The RABITQ path at line ~1787 has the same gap:

   ```cpp
   std::string index_file_path = FileHelper::MakeQuantizeVectorIndexPath(
       path_, column, segment_meta_->id(), quant_block_id);
   if (FileHelper::FileExists(index_file_path)) {
     LOG_WARN(
         "Index file[%s] already exists (possible crash residue); cleaning "
         "and overwriting.",
         index_file_path.c_str());
     FileHelper::RemoveFile(index_file_path);
   }
   auto vector_indexer = merge_vector_indexer(
       index_file_path, column, *field_with_new_index_params, concurrency);
   ```

   Apply the same pattern at both the non-RABITQ quantized path (lines 1715–1718) and the RABITQ path (lines 1787–1790).

7. `src/db/index/segment/segment.cc`, line 1639 ([link](https://github.com/alibaba/zvec/blob/b17d0de6747288b3973e503842fb2ff5898247cb/src/db/index/segment/segment.cc#L1639)) 

   **`RemoveFile` return value ignored in all four crash-recovery paths**

   `FileHelper::RemoveFile` returns a `bool` indicating success. If the removal fails (e.g., due to a permissions issue or the path being a directory), execution silently continues and `merge_vector_indexer` attempts to create the new index file while the stale, possibly-corrupted file is still present. This can cause the subsequent open/write to fail in confusing ways or produce a corrupt index — exactly the opposite of the recovery the code is trying to achieve.

   The existing code elsewhere in this file explicitly handles the failure case (e.g., line 3423):
   ```cpp
   if (!FileHelper::RemoveFile(filepath)) {
       return Status::InternalError("remove ", filepath, " failed");
   }
   ```

   The same guard is needed at all four new `RemoveFile` call sites:
   - Line 1639 (this hunk)
   - Line 1683
   - Line 1722
   - Line 1801

</details>

<sub>Reviews (5): Last reviewed commit: ["fix: fix by comments"](https://github.com/alibaba/zvec/commit/b17d0de6747288b3973e503842fb2ff5898247cb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=25810680)</sub>

### DIFF
--- a/src/db/index/segment/segment.cc
+++ b/src/db/index/segment/segment.cc
@@ -1714,6 +1714,13 @@ Status SegmentImpl::create_vector_index(
 
       std::string index_file_path = FileHelper::MakeQuantizeVectorIndexPath(
           path_, column, segment_meta_->id(), quant_block_id);
+      if (FileHelper::FileExists(index_file_path)) {
+        LOG_WARN(
+            "Index file[%s] already exists (possible crash residue); cleaning "
+            "and overwriting.",
+            index_file_path.c_str());
+        FileHelper::RemoveFile(index_file_path);
+      }
       auto vector_indexer = merge_vector_indexer(
           index_file_path, column, *field_with_new_index_params, concurrency);
       if (!vector_indexer.has_value()) {
@@ -1786,6 +1793,13 @@ Status SegmentImpl::create_vector_index(
 
       std::string index_file_path = FileHelper::MakeQuantizeVectorIndexPath(
           path_, column, segment_meta_->id(), quant_block_id);
+      if (FileHelper::FileExists(index_file_path)) {
+        LOG_WARN(
+            "Index file[%s] already exists (possible crash residue); cleaning "
+            "and overwriting.",
+            index_file_path.c_str());
+        FileHelper::RemoveFile(index_file_path);
+      }
       auto vector_indexer = merge_vector_indexer(
           index_file_path, column, *field_with_new_index_params, concurrency);
       if (!vector_indexer.has_value()) {


zhourrr:
@greptile

tests/db/crash_recovery/collection_optimizer.cc
```diff
@@ -0,0 +1,106 @@
...
+                 "directory: "
+              << e.what() << std::endl;
+  }
+
+  std::cout << "Configuration:" << std::endl;
+  std::cout << "  Path:        " << config.path << std::endl;
+  std::cout << std::endl;
+
+  auto result =
+      zvec::Collection::Open(config.path, zvec::CollectionOptions{false, true});
```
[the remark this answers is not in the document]
zhourrr: intensional

zhourrr:
@greptile

tests/db/crash_recovery/optimize_recovery_test.cc
```diff
@@ -0,0 +1,229 @@
...
+      for (auto &s : write_result.value()) {
+        ASSERT_TRUE(s.ok());
+      }
+    }
+
+    ASSERT_EQ(collection->Stats()->doc_count, num_batches * batch_size);
+    collection.reset();
+  }
+
+  RunOptimizerAndCrash(dir_path_, 4);
```
chinaux, line 132: 这个4秒怎么定的？ 会不会在有的机器optimize 执行的快，3秒就结束了？

tests/db/crash_recovery/collection_optimizer.cc
```diff
@@ -0,0 +1,106 @@
...
+
+  auto s = collection->Optimize();
+  if (s.ok()) {
+    std::cout << "Optimize completed successfully" << std::endl;
+    // Print final stats
+    std::cout << "Final stats: " << collection->Stats()->to_string_formatted() << std::endl;
+    return 0;
+  } else {
+    std::cout << "Optimize failed: " << s.message() << std::endl;
+    return 1;
```
chinaux, line 104: return -1 ?

tests/db/crash_recovery/optimize_recovery_test.cc
```diff
@@ -0,0 +1,229 @@
...
+  RunOptimizer(dir_path_);
+
+  // Open the collection and verify data integrity
+  auto result = Collection::Open(dir_path_, options_);
+  ASSERT_TRUE(result.has_value()) << "Failed to reopen collection after crash. "
+                                     "Recovery mechanism may be broken.";
+  auto collection = result.value();
+  uint64_t doc_count{collection->Stats().value().doc_count};
+  ASSERT_EQ(doc_count, (num_batches + 500) * batch_size);
+  for (uint64_t doc_id = 0; doc_id < doc_count; doc_id++) {
```
chinaux, line 197: 上面也有类似的逻辑，建议抽取为辅助函数或者加点注释

tests/db/crash_recovery/collection_optimizer.cc
zhourrr: This is meant to run in bash environment, where error code 1 means catchall for general errors

tests/db/crash_recovery/optimize_recovery_test.cc
zhourrr: 在所有ci上都测过了，4秒时间是足够的，实际过程都会超过4秒

zhourrr:
@greptile

tests/db/crash_recovery/optimize_recovery_test.cc
zhourrr: 这里没有抽象成函数主要是为了让报错日志时可以带上正确的 line number，不然可能会被嵌套在函数栈内，不清晰

zhourrr:
@greptile

PR #258, opened 2026-03-24
Refactor CPU feature detection to use an explicit x86 whitelist

### LINKED ISSUE #244: [Feature]: Add CI pipeline for RISC-V / RVV
### Problem / Motivation

As zvec involves performance-critical vector operations, supporting RISC-V Vector Extensions is a highly valuable capability. However, without a dedicated CI pipeline for the RISC-V architecture, any new commits to the C++ core might inadvertently break RVV optimizations, cause compilation failures on RISC-V targets, or introduce silent logic regressions.
Relying on developers to manually cross-compile and test RVV implementations locally before releases is prone to human error and slows down the development cycle. Having an automated CI workflow ensures that RVV code paths are continuously verified.

### Proposed Solution

```python

```

### Alternatives Considered

_No response_

### Affected Area

_No response_

ihb2032:
Currently, cpu_features.cc assumes that any architecture that is not __ARM_ARCH must be an x86/x64 architecture and attempts to include <cpuid.h>. This causes fatal compilation errors on other architectures like RISC-V.
Changes Proposed
This PR refactors the preprocessor directives to use an explicit "whitelist" approach:
  - <cpuid.h> and x86-specific CPUID logic are now only compiled when __x86_64__ or __i386__ is explicitly defined or under MSVC for Windows.
  - All other architectures including ARM, RISC-V, etc. will safely fall back to the default initialization, allowing the project to compile successfully as generic C++ code.

Motivation
This resolves build failures on RISC-V environments and sets up a clean, safe foundation for potentially adding architecture-specific optimizations in the future.
closed #244

### DIFF
--- a/.github/workflows/03-macos-linux-build.yml
+++ b/.github/workflows/03-macos-linux-build.yml
@@ -88,7 +88,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip \
+          python -m pip install --upgrade pip
+          python -m pip install \
             pybind11==3.0 \
             cmake==3.30.0 \
             ninja==1.11.1 \
--- a/.github/workflows/03-macos-linux-build.yml
+++ b/.github/workflows/03-macos-linux-build.yml
@@ -99,6 +99,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+<<<<<<< HEAD
           $PYTHON -m pip install --upgrade pip \
             pybind11==3.0 \
             cmake==3.30.0 \
@@ -107,6 +108,17 @@ jobs:
             pytest-xdist \
             scikit-build-core \
             setuptools_scm
+=======
+          $PYTHON -m pip install --upgrade pip          
+          DEPS="pybind11==3.0 cmake==3.30.0 ninja==1.11.1 pytest scikit-build-core setuptools_scm"
+          
+          if [[ "${{ inputs.os }}" == *"riscv"* ]]; then
+            echo "RISC-V architecture detected. Adding build tools for compiling numpy from source..."
+            DEPS="$DEPS meson-python meson Cython"
+          fi
+          
+          $PYTHON -m pip install $DEPS
+>>>>>>> 61d8516 (ci: add RISC-V numpy dependencies)
         shell: bash
 
       - name: Build from source
--- a/.github/workflows/03-macos-linux-build.yml
+++ b/.github/workflows/03-macos-linux-build.yml
@@ -97,6 +97,22 @@ jobs:
           echo "$($PYTHON -c 'import site; print(site.USER_BASE)')/bin" >> $GITHUB_PATH
         shell: bash
 
+      - name: Get pip cache dir
+        if: ${{ contains(inputs.os, 'riscv') }}
+        id: pip-cache
+        run: |
+          echo "dir=$($PYTHON -m pip cache dir)" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Cache pip dependencies for RISC-V
+        if: ${{ contains(inputs.os, 'riscv') }}
+        uses: actions/cache@v4
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-riscv-pip-${{ hashFiles('**/pyproject.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-riscv-pip-
+
       - name: Install dependencies
         run: |
 <<<<<<< HEAD
--- a/.github/workflows/03-macos-linux-build.yml
+++ b/.github/workflows/03-macos-linux-build.yml
@@ -97,22 +97,6 @@ jobs:
           echo "$($PYTHON -c 'import site; print(site.USER_BASE)')/bin" >> $GITHUB_PATH
         shell: bash
 
-      - name: Get pip cache dir
-        if: ${{ contains(inputs.os, 'riscv') }}
-        id: pip-cache
-        run: |
-          echo "dir=$($PYTHON -m pip cache dir)" >> $GITHUB_OUTPUT
-        shell: bash
-
-      - name: Cache pip dependencies for RISC-V
-        if: ${{ contains(inputs.os, 'riscv') }}
-        uses: actions/cache@v4
-        with:
-          path: ${{ steps.pip-cache.outputs.dir }}
-          key: ${{ runner.os }}-riscv-pip-${{ hashFiles('**/pyproject.toml') }}
-          restore-keys: |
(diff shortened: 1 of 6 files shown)


richyreachy:
hi @ihb2032, as we noticed that RISC-V is available on github runner (https://github.com/alibaba/zvec/issues/244), could you please setup a CI job on it along with this PR?

.github/workflows/03-macos-linux-build.yml
```diff
@@ -4,24 +4,23 @@ on:
   workflow_call:
```
richyreachy, line 4: as we checked, it's better to add one entry in **01-ci-pipeline.yml.** Takes clang as example, change os and platform to RISCV related. 


build-and-test-linux-x64-clang:
  name: Build & Test (linux-x64-clang)
  needs: lint
  uses: ./.github/workflows/03-macos-linux-build.yml
  with:
    platform: linux-x64-clang
    os: ubuntu-24.04
    compiler: clang

richyreachy: Otherwise, it would bring a bundle of CI jobs.

.github/workflows/03-macos-linux-build.yml
```diff
@@ -125,17 +124,3 @@ jobs:
           ./core-example
           ./ailego-example
         shell: bash
-
-      - name: Run C Examples
```
richyreachy, line 140: Hi, C examples job cannot to be deleted as used to test Zvec C API. Is there any concerns about it?

richyreachy: This file, as I can see, would not be changed.

richyreachy:
@ihb2032 please take a look at this error: https://github.com/alibaba/zvec/actions/runs/24500696519/job/71607486150

Seems the ci job can be brought up after Github App was setup. 

Any problem, let me know.

ihb2032:
Hi @richyreachy, thanks for pointing that out!

I checked the logs and the issue is that `actions/setup-python` currently doesn't provide a pre-built Python 3.10 binary for the `riscv64` architecture on Ubuntu 24.04.

To fix this without breaking the other builds that share `03-macos-linux-build.yml`, I've updated the workflow to conditionally use Python 3.12 only when `inputs.platform == 'linux-riscv64'`, while keeping the rest on 3.10.

I've just pushed the fix, let's see if the CI turns green!

ihb2032:
Hi @richyreachy, the RISC-V CI failed due to a timeout. Would it be possible to split the build and test phases to prevent this?

richyreachy:
> Hi @richyreachy, the RISC-V CI failed due to a timeout. Would it be possible to split the build and test phases to prevent this?

There's hard timeout limit(6 hours ) for CI jobs. Since test stage depends on the outcome of build, it may need some effects to figure out how to shrink the process.

ihb2032:
Hi @richyreachy,
To address the timeout while keeping the workflow stable, I'm thinking of a targeted optimization for the RISC-V build.
We could update the workflow to use the wheel provided by RISE for the cmake package, but only when the platform is linux-riscv64. For all other architectures, we’ll continue to use the standard packages from PyPI to ensure no regressions.
According to my estimate, this change could shave off nearly 1 hour from the RISC-V CI job. Does this conditional approach sound good to you?

richyreachy:
> Hi @richyreachy,
> 
> To address the timeout while keeping the workflow stable, I'm thinking of a targeted optimization for the RISC-V build.
> 
> We could update the workflow to use the wheel provided by RISE for the cmake package, but only when the platform is linux-riscv64. For all other architectures, we’ll continue to use the standard packages from PyPI to ensure no regressions.
> 
> According to my estimate, this change could shave off nearly 1 hour from the RISC-V CI job. Does this conditional approach sound good to you?

Yes. The workflow for RISC-V is slower compared to other architectures. Every measurement deserves a try for execution time optimization.

ihb2032:
Hi @richyreachy,
Found two failures in the CI:
1. C++ Tests: hnsw_streamer_test failed.
2. Python Tests: Failed due to the runner (rise-riscv-runner-8wj32esoq) missing the python3.10 environment (command not found).
The Python failure looks like a runner environment issue. Could you please help re-run the Python tests? Also, PTAL at the HNSW test failure.

richyreachy:
> Hi @richyreachy, Found two failures in the CI:
> 
> 1. C++ Tests: hnsw_streamer_test failed.
> 2. Python Tests: Failed due to the runner (rise-riscv-runner-8wj32esoq) missing the python3.10 environment (command not found).
>    The Python failure looks like a runner environment issue. Could you please help re-run the Python tests? Also, PTAL at the HNSW test failure.

The python test is still failing. 

@iaojnh  please take a look at the streamer test failure.  Looks like precision problem caused by randomly generated data.

ihb2032:
@richyreachy  Could you please re-run the failed Python and C++ test jobs? They failed due to a transient runner environment issue, not a code error. Thanks!

richyreachy:
> @richyreachy Could you please re-run the failed Python and C++ test jobs? They failed due to a transient runner environment issue, not a code error. Thanks!

Done. please keep tabs on them.

ihb2032:
@richyreachy Good news is that all the other RISC-V tests have passed. As I pointed out earlier, the C++ test failure is just down to the hnsw_streamer_test case.

Glad to hear it! Could @iaojnh please take a look at the streamer test failure? To my knowledge, it should already be resolved.

Here's the fix: https://github.com/alibaba/zvec/pull/375

ihb2032:
@richyreachy Looks like the remaining C++ test failure is still stemming from hnsw_streamer_test.

ihb2032:
@richyreachy @iaojnh It looks like the CI is still failing on hnsw_streamer_test. Could you take another look?

ihb2032:
Hi @iaojnh, FYI, the linux-riscv64 C++ CI is failing on `HnswStreamerTest.TestKnnSearchCosine` due to a borderline recall assertion.

richyreachy:
hi @ihb2032 can you use cache by following this commit? https://github.com/alibaba/zvec/commit/d91867e00edb59a6fdc9a63fabd0b6db5f0d699d

richyreachy:
hi @ihb2032 

Good news that all runtime errors have been fixed!

As we discussed internally, since the workflow runs quite slowly, it would be more efficient to run this CI job nightly instead. Here’s an example for reference:
https://github.com/alibaba/zvec/blob/main/.github/workflows/nightly_coverage.yml

Any problem or concern, please let me know.

ihb2032:
Makes sense! Thanks for the reference link. I will update the workflow trigger to run nightly and push the update.

richyreachy (approved):
LGTM

PR #282, opened 2026-03-30
fix: keep index type enum value for compatibility

egolearner:


src/include/zvec/db/type.h
```diff
@@ -20,12 +20,13 @@ namespace zvec {
 /*
  * Column index types
  */
+// NOTE: enum values are not the same as zvec.proto
```
zhourrr, line 23: 这个 comment 可以写一下不一样的原因，是bug，不是feature，让读者更好理解这个处境

egolearner: Closed since compatibility is not an issue.

egolearner:
Closed since compatibility is not an issue.

PR #297, opened 2026-04-01
fix: remove redundant Python-side vector index type validation in create_index

songqing:
The original check used `in` operator to validate whether an index param
instance belongs to vector index types, which is always False since it
compares an instance against classes. Rather than fixing it with
`isinstance`, remove the entire Python-side validation and delegate type
checking to the C++ core, which already handles it properly and avoids
the risk of the Python type list falling out of sync.

### DIFF
--- a/python/zvec/model/collection.py
+++ b/python/zvec/model/collection.py
@@ -39,13 +39,6 @@ from .schema import CollectionSchema, CollectionStats, FieldSchema
 
 __all__ = ["Collection"]
 
-_VECTOR_INDEX_TYPES = (
-    HnswIndexParam,
-    HnswRabitqIndexParam,
-    IVFIndexParam,
-    FlatIndexParam,
-)
-
 
 class Collection:
     """Represents an opened collection in Zvec.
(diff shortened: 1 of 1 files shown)


zhourrr (approved):

python/zvec/model/collection.py
```diff
@@ -136,7 +136,7 @@ def create_index(
         Raises:
             ValueError: If a vector index is applied to a non-vector field.
         """
-        if index_param in _VECTOR_INDEX_TYPES and not self.schema.vector(field_name):
+        if isinstance(index_param, _VECTOR_INDEX_TYPES) and not self.schema.vector(field_name):
```
zhourrr: Thank you for this catch, this validation was indeed broken before. One minor thing: this file needs to go through ruff format. Could you please run:
"ruff format python/zvec/model/collection.py" and push the updated formatting?

zhourrr (changes requested):

python/zvec/model/collection.py
songqing: @zhourrr done

zhourrr: @songqing After looking closer, this check on python side is redundant because the C++ core code handles the validation. Could you please just delete this check (line 139-144 in the original version)? That should resolve the issue. Thanks!

songqing: done

zhourrr (approved):

PR #314, opened 2026-04-03
test: add test_collection_ddl  cases

liututu12:
1.Create index functionality to supplement support for all data types, index methods, and metric combinations, as well as       unsupported scenarios;
    2.Supplement test cases for the collection.optimize method;
    3.After any DDL operation (creating an index, adding/modifying/deleting a column, collection.flush(), optimize(),
    etc.), close the collection, reopen it, and write data to check if this process works correctly.


python/tests/detail/params_helper.py
```diff
@@ -36,9 +36,9 @@
             quantize_type=QuantizeType.FP16,
         ),
         FlatIndexParam(),
-        FlatIndexParam(metric_type=MetricType.IP, quantize_type=QuantizeType.INT4),
-        FlatIndexParam(metric_type=MetricType.L2, quantize_type=QuantizeType.INT8),
-        FlatIndexParam(metric_type=MetricType.COSINE, quantize_type=QuantizeType.FP16),
```
feihongxu0824, line 41: why remove this case?

python/tests/detail/test_collection_ddl.py
```diff
@@ -16,6 +16,128 @@
 from fixture_helper import *
 from doc_helper import *
 from params_helper import *
+import threading, time
+
+indextest_collection_schema = zvec.CollectionSchema(
```
feihongxu0824, line 21: why not use full_schema in fixture_helper?

python/tests/detail/test_collection_ddl.py
```diff
@@ -16,6 +16,128 @@
...
+            "sparse_fp32_field",
+            DataType.SPARSE_VECTOR_FP32,
+            dimension=128,
+            index_param=HnswIndexParam(),
+        ),
+    ],
+)
+
+
+def batchdoc_and_check(collection: Collection, multiple_docs, operator="insert"):
```
feihongxu0824, line 103: Why are there so many places with batchdoc_and_check? Why not unify them?

PR #329, opened 2026-04-08
fix(quantizer): use rounded int8 values for SQ8 metadata to fix recall drop

### LINKED ISSUE #328: [Bug]: SQ8 quantization causes significant recall drop on OpenAI Performance1536D50K dataset
### Description

When benchmarking zvec v0.2.1 with SQ8 quantization on the OpenAI Performance1536D50K dataset using VectorDBBench, I observed a significant recall drop to 0.7377, which is much lower than expected.

Command used:

```
vectordbbench zvec --path Performance1536D50K --db-label 16c64g-v0.1 \
    --case-type Performance1536D50K --num-concurrency 1 \
    --quantize-type int8 --m 15 --ef-search 180
```

After investigating, I found that the root cause is in `src/core/quantizer/record_quantizer.h`. Currently, the metadata fields sum (extras[2]) and squared_sum (extras[3]) are computed from the pre-rounded float values to preserve precision. However, on datasets with asymmetric value ranges — like OpenAI embeddings, where |x_min| ≈ 0.64 >> x_max ≈ 0.21, causing most values to be compressed into the right half of the quantization interval [−127, 127] — computing these metadata fields from the rounded int8 values instead actually yields significantly better recall.

By changing the computation to use rounded values — consistent with what is actually stored — recall improves from 0.7377 to 0.9588 on the same benchmark, with no regression observed on other datasets (Cohere 768D, BIOASQ 1024D).

A detailed mathematical analysis and experimental validation are included below. The document was collaboratively authored by me and Claude Code, and is written in Chinese.

[sq8_recall_analysis.md](https://github.com/user-attachments/files/26567781/sq8_recall_analysis.md)

### Steps to Reproduce

```python
1. pip install zvec==v0.2.1
2. vectordbbench zvec --path Performance1536D50K --db-label 16c64g-v0.1 \
    --case-type Performance1536D50K --num-concurrency 1 \
    --quantize-type int8 --m 15 --ef-search 180
```

### Logs / Stack Trace

```shell

```

### Operating System

Ubuntu 22.04

Python 3.11

### Additional Context

JoeJRW:
fixes #328

Problem:
SQ8 metadata (squared_sum, sum) was computed from pre-rounded float values, causing mismatch with actual stored int8 values. On asymmetric datasets (e.g. OpenAI 1536D where |x_min| >> x_max), this leads to severe recall drop.

Solution:
Move std::round before accumulating squared_sum and sum.

### DIFF
--- a/src/core/quantizer/record_quantizer.h
+++ b/src/core/quantizer/record_quantizer.h
@@ -44,11 +44,10 @@ class RecordQuantizer {
         scale = 254 / std::max(max - min, epsilon);
         bias = -min * scale - 127;
         for (size_t i = 0; i < dim; ++i) {
-          float v = vec[i] * scale + bias;
+          float v = std::round(vec[i] * scale + bias);
           squared_sum += v * v;
           sum += v;
-          (reinterpret_cast<int8_t *>(out))[i] =
-              static_cast<int8_t>(std::round(v));
+          (reinterpret_cast<int8_t *>(out))[i] = static_cast<int8_t>(v);
           int8_sum += (reinterpret_cast<int8_t *>(out))[i];
         }
         extras = reinterpret_cast<float *>(static_cast<int8_t *>(out) + dim);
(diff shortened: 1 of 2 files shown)


iaojnh:
@JoeJRW Thank you very much for this contribution! We need time to assess the impact of this change. Could you provide some test data to support your point?

JoeJRW:
@iaojnh You can take a look at my tests and analysis in the issue #328. I've pasted my test results in the markdown document. I used the vectordbbench tool to test openai1536D, cohere768D, and bioasq1024D. You can try testing and reproducing it.

richyreachy (dismissed):

richyreachy (approved):
LGTM

PR #338, opened 2026-04-14
feat: make max query topk configurable

Rm1n90:
## Summary

- Adds a `max_query_topk` parameter to `zvec.init()` so users can raise the hard topk ceiling (default 1024) when they accept the extra memory/latency tradeoff
- Stores the configured limit in `GlobalConfig` and reads it in `VectorQuery::validate()` instead of using the hardcoded `kMaxQueryTopk` constant
- Exposes the same config through the C API via `zvec_config_data_{set,get}_max_query_topk()`

### Usage

```python
import zvec

zvec.init(max_query_topk=4096)
results = collection.query(vq, topk=2048)  # now works
```

```c
zvec_config_data_t *cfg = zvec_config_data_create();
zvec_config_data_set_max_query_topk(cfg, 4096);
zvec_initialize(cfg);
```

### Files changed (7)

| Layer | File | Change |
|---|---|---|
| C++ core | `src/include/zvec/db/config.h` | Add `max_query_topk` to `ConfigData` + accessor |
| C++ core | `src/db/common/config.cc` | Default value (1024) + validation |
| C++ core | `src/db/index/common/doc.cc` | Use configurable limit in `VectorQuery::validate()` |
| Python | `python/zvec/zvec.py` | New `max_query_topk` param with type/range validation |
| Python binding | `src/binding/python/model/common/python_config.cc` | Parse and validate from config dict |
| C API | `src/include/zvec/c_api.h` | Declare getter/setter |
| C API | `src/binding/c/c_api.cc` | Implement getter/setter |

## Test plan

- [x] Full build from source (macOS ARM64, Python 3.13)
- [x] Python test suite: **1009 passed**, 54 skipped, 0 failures
- [x] Validated edge cases: float, negative, zero, bool, string all correctly rejected with clear error messages
- [x] Default behavior unchanged — limit stays 1024 unless explicitly overridden

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/python/zvec/zvec.py
+++ b/python/zvec/zvec.py
@@ -39,6 +39,7 @@ def init(
     invert_to_forward_scan_ratio: Optional[float] = None,
     brute_force_by_keys_ratio: Optional[float] = None,
     memory_limit_mb: Optional[int] = None,
+    max_query_topk: Optional[int] = None,
 ) -> None:
     """Initialize Zvec with configuration options.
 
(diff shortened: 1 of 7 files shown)


zhourrr:
This issue has been resolved in #348. While the topk limit has been increased, it was not exposed as a configurable parameter to maintain simplicity. We hope this new limit will accommodate most use cases.

PR #347, opened 2026-04-17
minor: fix c_api cmake

feihongxu0824:
1. remove useless dependancies in c example CMake;
2. build a c++ fat dynamic lib and update c++ example CMake.


chinaux (approved):
lgtm

PR #363, opened 2026-04-21
enhance: release gil lock in collection bindings

feihongxu0824:


src/binding/python/model/python_collection.cc
```diff
@@ -75,13 +75,20 @@ void ZVecPyCollection::bind_db_methods(
   col.def_static("CreateAndOpen",
                  [](const std::string &path, const CollectionSchema &schema,
                     const CollectionOptions &options) {
-                   auto result =
-                       Collection::CreateAndOpen(path, schema, options);
+                   Result<Collection::Ptr> result;
```
egolearner, line 78: Minor: DRY, use template+lambda

```
template <typename Func>
auto call_without_gil(Func &&func) {
  using R = std::invoke_result_t<Func>;
  R result;
  {
    py::gil_scoped_release release;
    result = std::forward<Func>(func)();
  }
  return result;
}
```

```
auto result = call_without_gil([&] { return self.Insert(docs); });
return unwrap_expected(result);
```

egolearner (dismissed):

egolearner (approved):

PR #378, opened 2026-04-29
feat/diskann dynamic load

richyreachy:
Use dynamic loading to avoid environmental dependencies.


richyreachy:
cpp usage would be like: 
```
#include <zvec/plugin/diskann_plugin.h>

if (zvec::IsLibAioAvailable()) {
    if (zvec::LoadDiskAnnPlugin() == zvec::kDiskAnnPluginOk) {
        // Now safe to create DiskAnnIndex / DiskAnnBuilder / DiskAnnStreamer.
    }
}
```

python usage would be like:
```
import zvec

if zvec.is_libaio_available():
    status = zvec.load_diskann_plugin()          
    assert status == zvec.DiskAnnPluginStatus.OK
  
```

PR #382, opened 2026-04-30
fix: sparse vector indices should be ordered

### LINKED ISSUE #379: [Bug]: Sparse vector search returns different results based on element insertion order
### Description

The sparse vector search returns completely different results depending on the order in which key-value pairs are ordered in the sparse vector dictionary. The same vector with reordered elements produces different scores and rankings.

```python
# Same vector, different orderings
sorted_vec   = {1: 0.5, 5: 1.2, 10: 0.8, 1000: 3.0}
reversed_vec = {1000: 3.0, 10: 0.8, 5: 1.2, 1: 0.5}
```

Possible fix: 
Make sure vectors are sorted during insertion/query
Improve the documentation

### Steps to Reproduce

```python
1. Index documents with sparse vectors
2. Query with the same query vector
3. Compare results when sparse vectors are:
     - Sorted (default)
     - Reversed (keys in reverse order)
     - Shuffled (random order)
```

### Logs / Stack Trace

```shell

```

### Operating System

Ubuntu 24.04 LTS

Python 3.12

### Additional Context

zhourrr:
This should fix #379

### DIFF
--- a/src/binding/python/model/param/python_param.cc
+++ b/src/binding/python/model/param/python_param.cc
@@ -97,6 +97,9 @@ std::pair<std::string, std::string> serialize_sparse_vector(
   const size_t n = sparse_dict.size();
   if (n == 0) return {{}, {}};
 
+  const auto sorted_items = py::module_::import("builtins")
+                                .attr("sorted")(sparse_dict.attr("items")());
+
   std::string indices_buf;
   indices_buf.resize(n * sizeof(uint32_t));
   auto *indices_ptr = reinterpret_cast<uint32_t *>(indices_buf.data());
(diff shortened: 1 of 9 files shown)


zhourrr:
@codex review

JalinWang:
@codex review

src/binding/python/model/param/python_param.cc
```diff
@@ -97,6 +97,9 @@ std::pair<std::string, std::string> serialize_sparse_vector(
   const size_t n = sparse_dict.size();
   if (n == 0) return {{}, {}};
 
+  const auto sorted_items = py::module_::import("builtins")
```
chinaux: It's better to sort at the C++ layer; otherwise, we'd also need to modify the extract_sparse_vector function in the C API.

tests/db/index/common/doc_test.cc
```diff
@@ -668,26 +669,52 @@ TEST_F(DocDetailedTest, Validate) {
     ASSERT_EQ(s.code(), StatusCode::INVALID_ARGUMENT);
   }
 
-  // doc validate error
+  // sparse vector indices must be strictly ascending and unique
+  {
+    auto schema = test::TestHelper::CreateNormalSchema(false);
+    auto doc = test::TestHelper::CreateDoc(1, *schema);
+
+    // unsorted indices are rejected
```
chinaux: Need to add two more test cases:
1.One-dimensional input
2.Inputs with different sizes

src/binding/python/model/param/python_param.cc
chinaux: It would be better to sort directly in C++ here— the current approach incurs overhead from crossing the C++/Python boundary.

tests/db/index/common/doc_test.cc
zhourrr: one dimensional input is supported and will skip check.

Inputs with different sizes is tested in previous test on line 645

zhourrr:
@codex review

chinaux (approved):
lgtm

Issue #199, opened 2026-03-05
[Feature]: Support fts (full text search)


sandsc:
### Problem / Motivation

How do I create a multilingual FTS index and perform hybrid searches that combine full-text and vector queries?

### Proposed Solution

```python

```

egolearner:
FTS is a planned feature currently under design and implementation. It is scheduled to be supported in version 0.4.x, with an estimated release this summer.

sandsc:
Nice job! It would be wonderful if it could support Chinese directly.

paulocoutinhox:
This is the best of two worlds in a single library. 

I want use it inside IonClaw https://github.com/ionclaw-org/ionclaw.

PR #408, opened 2026-05-15
feat: add fts support

egolearner:
address #397 #398 #399 #199


python/tests/test_fts_query.py
```diff
@@ -0,0 +1,158 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for FTS (Full-Text Search) query support in the Python SDK."""
+
+import pickle
+
+import pytest
+
+from zvec.model.param.query import Fts, Query
```
JalinWang, line 20: This naming "Fts" is a little bit too generic. Would it be more precise to name it after its underlying dependency, like _FtsQuery (binding) or FtsQueryParam (C++)?

egolearner: The current design is: `Query` is the top-level request containing multiple retrieval fields (vector, FTS, filter, etc.), and `Fts` is just one component within it — similar to how you'd have `Vector` or `Filter` as sibling fields.

Naming it `FtsQuery` would imply it's a standalone query type at the same level as `Query`, which is misleading. If we later split `Query` into multiple specialized types, `FtsQuery` would then be the right name for a full FTS query request.

We'll also align the C++ side to drop the "Query" suffix from the param struct for consistency.

What do you think? Any other naming suggestions?

JalinWang: You’re right — FtsQuery doesn’t feel quite right.

That said, Fts alone also feels a bit awkward, since it doesn’t make it clear that this is a component of Query, not a standalone type.

Maybe we should consider a name that better reflects that role, such as FtsQueryParam, FtsClause, FtsCondition, FtsOption, or FtsFilter.

python/tests/test_fts_query.py
```diff
@@ -0,0 +1,158 @@
...
+        """Query without FTS fields should have has_fts() == False."""
+        q = Query(field_name="embedding", vector=[0.1, 0.2, 0.3])
+        assert q.has_fts() is False
+
+    def test_vector_and_fts_mutually_exclusive(self):
+        """Cannot combine vector search with FTS in a single Query."""
+        q = Query(
+            field_name="embedding",
+            vector=[0.1, 0.2, 0.3],
+            fts=Fts(match_string="deep learning"),
```
JalinWang, line 63: Also, would it be better to rename `fts=` here to something more explicit, like `full_text=` / `full_text_search=` / `text_search=`? It feels a bit clearer and makes the intent more obvious, especially for users who may not immediately recognize fts as full-text search.

src/db/sqlengine/parser/select_info.h
```diff
@@ -82,6 +95,7 @@ class SelectInfo : public BaseInfo {
   int limit_{-1};
   bool include_vector_{false};
   bool include_doc_id_{false};
+  FtsCondInfo::Ptr fts_cond_info_{nullptr};
```
feihongxu0824, line 98: select_info的拷贝构造和赋值运算符中没有拷贝fts cond，include_vector、include_doc_id、groupby也没有

feihongxu0824: 这里是不是缺少命中sql cache的单测

src/db/sqlengine/planner/query_planner.cc
```diff
@@ -616,6 +628,28 @@ Result<PlanInfo::Ptr> QueryPlanner::invert_scan(
...
+  auto source_node_options =
+      arrow::acero::SourceNodeOptions{recall_node->schema(), recall_node->gen(),
+                                      arrow::compute::Ordering::Implicit()};
+  ac::Declaration node{"source", source_node_options};
+
+  node = ac::Declaration{
+      "fetch", {std::move(node)}, ac::FetchNodeOptions{0, topn}};
+  return std::make_shared<PlanInfo>(std::move(node), recall_node->schema());
+}
+
```
feihongxu0824, line 661: 多个segment的fts计划里面没有order_by节点，这个符合预期吗？

src/db/index/column/fts_column/iterator/fts_term_iterator.cc
```diff
@@ -0,0 +1,203 @@
...
+
+// ============================================================
+// Block-Max WAND support
+// ============================================================
+
+DocIterator::BlockMaxInfo TermDocIterator::block_max_info_for(
+    uint32_t target) const {
+  if (mode_ == Mode::BITPACKED) {
+    auto info = bp_iter_.block_max_info_for(target);
+    return {info.block_max_score, info.block_last_doc};
```
feihongxu0824: 这里返回的是不是也需要乘上boost？

src/db/index/segment/segment.cc
```diff
@@ -294,6 +306,12 @@ class SegmentImpl : public Segment,
 
   Status reopen_invert_indexer(bool read_only = false);
 
+  // FTS helpers
+  Status open_fts_indexers(bool create);
+  Status close_fts_indexers();
+  Status flush_fts_indexers();
```
feihongxu0824, line 316: 这一期暂时不支持create/drop fts index吗？

src/db/proto/zvec.proto
```diff
@@ -131,6 +133,12 @@ message VamanaIndexParams {
   bool use_id_map = 7;
 }
 
+message FtsIndexParams {
+  string tokenizer_name = 1;
+  repeated string filters = 2;
+  string extra_params = 3;
```
feihongxu0824, line 139: extra_params暂时没用到？

src/include/zvec/db/index_params.h
```diff
@@ -558,4 +565,98 @@ class VamanaIndexParams : public VectorIndexParams {
...
+    }
+    auto &other_fts = static_cast<const FtsIndexParams &>(other);
+    return tokenizer_name_ == other_fts.tokenizer_name_ &&
+           filters_ == other_fts.filters_ &&
+           extra_params_ == other_fts.extra_params_;
+  }
+
+  //! Thread-safe lazy creation of TokenizerPipeline.
+  //! Returns the cached pipeline on subsequent calls.
+  Result<PipelinePtr> create_pipeline();
```
feihongxu0824: 这个接口只是内部用的，没必要透出给用户

src/db/index/column/fts_column/iterator/fts_term_iterator.cc
egolearner: 是的，已修复

src/db/index/segment/segment.cc
egolearner: 要不后面再添加支持吧，当前PR已经很大了。先只支持schema中预先创建好fts字段。

src/db/proto/zvec.proto
egolearner: jieba分词用到了

src/db/sqlengine/parser/select_info.h
egolearner: 拷贝构造和赋值运算符没用到，但是cache逻辑有用到，后面清理的时候改为 = delete吧

sql info的cache没用到，只用到了filter的cache。后面可以把sql info的cache去掉

src/db/sqlengine/planner/query_planner.cc
egolearner: 已修复和增加ut。

python/tests/test_fts_query.py
egolearner: Appreciate the suggestion! I get that full_text= is more explicit. But fts is shorter and well-understood in this space, so I'd lean toward keeping it. Thanks for understanding!

BTW, sqlite simply uses `fts5`
```sql
CREATE VIRTUAL TABLE email USING fts5(sender, title, body);
```

python/tests/test_fts_query.py
egolearner: We're adding `VectorClause` and `FtsClause` in the C++ SDK.
https://github.com/alibaba/zvec/pull/405/changes#diff-3c93a0b897a78e57cd5990f965fd452e6fc52765d8c93366ec47182fab03f2efR63

However, the Python SDK follows a more concise style. Since there's no clause abstraction in other Python scenarios, it would feel inconsistent to introduce FtsClause alone.

src/include/zvec/db/index_params.h
egolearner: 已隐藏

src/db/index/segment/segment.cc
feihongxu0824: 如果用户调用会获得错误信息的逻辑需要加上

egolearner: 已完善CreateIndex/DropIndex校验逻辑和补充ut

src/db/common/rocksdb_context.cc
```diff
@@ -112,134 +91,125 @@ Status RocksdbContext::create(
...
     }
   } else {
     bool has_default = false;
-    for (const auto &column_name : column_names) {
+    for (const auto &column_name : args.column_names) {
       if (std::find(existing_cf_names.begin(), existing_cf_names.end(),
                     column_name) == existing_cf_names.end()) {
-        LOG_ERROR("Column family[%s] does not exist in RocksDB[%s]",
-                  column_name.c_str(), db_path.c_str());
+        LOG_WARN("Column family[%s] does not exist in RocksDB[%s]",
```
zhourrr: 加入 fts 功能之后，会经常打出这句 warning 吗？因为我们设置的默认 log level 是 warning，用户可能会看到很多这种 warning log

feihongxu0824 (dismissed):
LGTM

src/db/common/rocksdb_context.cc
egolearner: fts字段在mutable会创建一些CF，dump时drop掉，因此需要检测CF是否存在，目前的做法是先用完整CF列表打开，如果打开失败再去掉这些CF再打开一次。所以可能会有warning

egolearner: 已改为使用ListColumnFamilies的结果，同时日志恢复ERROR

feihongxu0824 (approved):

PR #410, opened 2026-05-18
fix: nullable scalar field filter leaks null documents without inverted index

### LINKED ISSUE #409: [Bug] Nullable scalar field filter leaks null documents when no inverted index is present
## Description

When a scalar field is defined with `nullable=True` and **no inverted index** (`InvertIndexParam`) is created on it, query filter expressions incorrectly return documents where the field value is null. The null documents should be excluded by the filter but are returned alongside valid matches.

When an inverted index **is** created on the same field, the filter works correctly and null documents are properly excluded.

## Affected Types

**All 8 scalar types** are affected:

- `INT32`, `INT64`, `UINT32`, `UINT64`
- `FLOAT`, `DOUBLE`
- `STRING`, `BOOL`

## Relationship with Index

| Scenario | Result |
|---|---|
| **No inverted index** (`InvertIndexParam` not set) | **BUG** — null documents incorrectly returned |
| **With inverted index** (`InvertIndexParam` set) | **OK** — only matching non-null documents returned |

## Reproduction

**Environment:**
- zvec version: 0.4.0
- Python: 3.14
- OS: macOS (Apple Silicon)

**Minimal reproduction script:**

```python
import zvec
import numpy as np

DIM = 8

    name="test_null_bug",
    fields=[
        zvec.FieldSchema(
            name="val",
            data_type=zvec.DataType.INT64,
            nullable=True,
            # No index_param — this triggers the bug
        )
    ],
    vectors=[
        zvec.VectorSchema(
            name="emb",
            data_type=zvec.DataType.VECTOR_FP32,
            dimension=DIM,
        )
    ],
)

coll = zvec.create_and_open("./test_null_bug", schema)

np.random.seed(123)
docs = []
for i in range(3):
    docs.append(zvec.Doc(
        id=f"doc_{i}",
        vectors={"emb": np.random.rand(DIM).astype(np.float32).tolist()},
        fields={"val": 42},
    ))
for i in range(3, 6):
    docs.append(zvec.Doc(
        id=f"doc_{i}",
        vectors={"emb": np.random.rand(DIM).astype(np.float32).tolist()},
        fields={"val": None},
    ))

coll.insert(docs)
coll.flush()

results = coll.query(
    vectors=zvec.VectorQuery(field_name="emb", vector=np.random.rand(DIM).astype(np.float32).tolist()),
    topk=10,
    filter="val > 0",
    output_fields=["val"],
)

print(f"Expected: 3 docs (only non-null)")
print(f"Actual: {len(results)} docs")
for doc in results:
    print(f"  {doc.id} -> val={doc.field('val')}")

coll.destroy()
```

**Expected output:**
```
Expected: 3 docs (only non-null)
Actual: 3 docs
  doc_0 -> val=42
  doc_2 -> val=42
  doc_1 -> val=42
```

**Actual output:**
```
Expected: 3 docs (only non-null)
Actual: 6 docs
  doc_0 -> val=42
  doc_4 -> val={}
  doc_2 -> val=42
  doc_5 -> val={}
  doc_1 -> val=42
  doc_3 -> val={}
```

## Additional Notes

1. Null field values are returned as `{}` (empty dict) instead of `None` in the query results.
2. The bug likely exists in the brute-force scan filter path — when no inverted index is present, the null bitmap/flag is not being checked, causing null values to pass through the filter as if they satisfy the condition.
3. Adding `InvertIndexParam()` to the field schema or calling `create_index()` with `InvertIndexParam()` is a workaround, but the brute-force path should also correctly handle nulls.

feihongxu0824:
## Summary

- Fix null document leakage through filter when nullable scalar fields have no inverted index
- `get_forward_bit()`: Arrow's `BooleanArray::operator[]` returns `nullopt` for null entries, which `is_filtered()` treats as "no filter" — fix with `value_or(false)`
- `is_matched_by_forward_filter()`: reads `BooleanScalar.value` without checking `is_valid` (UB for null scalars) — fix by checking `is_valid` first

Closes #409

## Test plan

- [x] Added unit test `Feature_Query_NullableFilter_WithoutIndex`
- [x] Verified test **fails** without the fix (null docs leak through filter)
- [x] Verified test **passes** with the fix applied
- [x] Test covers both paths: without inverted index (buggy path) and with inverted index (control)

### DIFF
--- a/src/db/sqlengine/planner/doc_filter.cc
+++ b/src/db/sqlengine/planner/doc_filter.cc
@@ -98,7 +98,7 @@ std::optional<bool> DocFilter::get_forward_bit(uint64_t id) const {
     const auto &arr = forward_bitmap_->chunk(c);
     if (id < rows_seen + arr->length()) {
       auto *bool_array = static_cast<arrow::BooleanArray *>(arr.get());
-      return (*bool_array)[id - rows_seen];
+      return (*bool_array)[id - rows_seen].value_or(false);
     }
     rows_seen += arr->length();
   }
@@ -155,7 +155,8 @@ std::optional<bool> DocFilter::is_matched_by_forward_filter(uint64_t id) const {
   }
   arrow::Datum datum = maybe_result.MoveValueUnsafe();
   if (datum.is_scalar()) {
-    return datum.scalar_as<arrow::BooleanScalar>().value;
+    const auto &scalar = datum.scalar_as<arrow::BooleanScalar>();
+    return scalar.is_valid && scalar.value;
   }
   LOG_ERROR("Datum is not scalar, id[%zu] type[%s]", (size_t)id,
             datum.type()->ToString().c_str());
--- a/tests/db/collection_test.cc
+++ b/tests/db/collection_test.cc
@@ -4341,3 +4341,58 @@ TEST_F(CollectionTest, CornerCase_CreateIndex) {
   ASSERT_FALSE(s.ok());
   ASSERT_EQ(s.code(), StatusCode::INVALID_ARGUMENT);
 }
+
+TEST_F(CollectionTest, Feature_Query_NullableFilter_WithoutIndex) {
+  auto run_test = [&](bool with_scalar_index) {
+    FileHelper::RemoveDirectory(col_path);
+    IndexParams::Ptr scalar_idx =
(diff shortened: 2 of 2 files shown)


egolearner (approved):
LGTM

PR #413, opened 2026-05-18
build(thirdparty): mark third-party include paths as SYSTEM

egolearner:
Add SYSTEM to target_include_directories in vendored CMake wrappers and set INTERFACE_SYSTEM_INCLUDE_DIRECTORIES on imported / sub-project targets so warnings from third-party headers no longer surface in our build output. Introduces a mark_target_includes_system() helper in cmake/utils.cmake (resolves aliases, skips missing targets) used by the upstream-add_subdirectory wrappers (glog, gflags, googletest, yaml-cpp, protobuf, antlr).


chinaux (approved):
lgtm

PR #419, opened 2026-05-21
chore: enable modernize-use-override and fix existing violations

egolearner:
Add modernize-use-override to .clang-tidy and apply fixes across src/ and tests/: replace redundant virtual with override, annotate missing override on derived methods, and drop redundant virtual on already-overridden methods.


egolearner:
resolve #290

zhourrr (approved):

PR #423, opened 2026-05-26
minor: rewrite repeated optimize unit tests to cover more indexes

zhourrr:
# HNSW_RABITQ Repeated Optimize Bug — Fix Report

## Problem

`CollectionTest.Feature_Optimize_Repeated` failed for `HNSW_RABITQ`. Once Phase 4 inserted a single doc and triggered re-optimize, the new index file could not be opened:

```
Failed to get segment rabitq.converter
Failed to load reformer, ret=-32
Failed to open streamer, path: test_collection/0.tmp/dense_fp32.qindex.2.proxima
optimize failed: Failed to open index
```

## Root Cause

The single-segment incremental path in `SegmentImpl::create_vector_index` and the multi-segment compact path in `SegmentHelper::ReduceVectorIndex` were **asymmetric** for RaBitQ:

- The single-segment path trains a `RabitqConverter`, derives a reformer via `to_reformer()`, and attaches it to a cloned `HnswRabitqIndexParams` so the streamer can dump the `rabitq.converter` segment into the new index file.
- The compact path skipped these steps and constructed the quantize `VectorColumnIndexer` directly from the original schema field (no reformer attached).

When the test inserted one doc, two persisted segments existed → `build_compact_task` chose `CompactTask` → `ReduceVectorIndex` ran. With no reformer on the field, `HnswRabitqStreamer::open` entered the `reformer_ == nullptr` branch and tried to `load()` the `rabitq.converter` segment from a brand-new (empty) storage, which failed and aborted optimize.

## Fix

In `src/db/index/segment/segment_helper.cc`, within the quantize branch of `ReduceVectorIndex`:

1. After flushing the flat indexer, for `quantize_type == RABITQ` (gated by `#if RABITQ_SUPPORTED`):
   - Obtain `raw_vector_provider` from the still-open flat indexer.
   - Create and `init` a `RabitqConverter` with the same metric/dimension/RABITQ params used by the single-segment path.
   - `train(raw_vector_provider)` and `to_reformer(&reformer)`.
   - Clone `HnswRabitqIndexParams`, call `set_rabitq_reformer(reformer)` and `set_raw_vector_provider(raw_vector_provider)`, and put the cloned params onto a `field_for_quantize`.
2. Build the quantize `VectorColumnIndexer` using `field_for_quantize` (instead of the original `*field`), so the reformer flows through `engine_helper.hpp` → `HNSWRabitqIndex::CreateAndInitStreamer` → `HnswRabitqStreamer::open`, where the `else` branch correctly dumps `rabitq.converter` into the storage.
3. Defer `vector_indexer->Close()` on the flat indexer until **after** the quantize index is built, so the `raw_vector_provider` it backs stays valid throughout converter training and quantize index construction.
4. Added the necessary includes (`rabitq_params.h` under `RABITQ_SUPPORTED`, `engine_helper.hpp`, `zvec/core/interface/index.h`, and the relevant core framework headers).

Non-RABITQ quantize types are unaffected — `field_for_quantize` is simply a copy of `*field` for them.

## Verification

```
[ RUN      ] CollectionTest.Feature_Optimize_Repeated
[       OK ] CollectionTest.Feature_Optimize_Repeated (114515 ms)
[  PASSED  ] 1 test.
```

The repeated `FhtKacRotator is selected` log lines confirm the converter is now trained on every compact iteration, exactly as designed.

## Risk

- Change is local to a single function plus a header-include block.
- No on-disk format, proto, or public API change.
- Pattern mirrors a proven implementation already in `SegmentImpl::create_vector_index`.
- Other quantize types and the single-segment optimize path are untouched.

### DIFF
--- a/src/db/index/segment/segment_helper.cc
+++ b/src/db/index/segment/segment_helper.cc
@@ -18,29 +18,18 @@
 #include <memory>
 #include <arrow/compute/api_vector.h>
 #include <arrow/type_fwd.h>
-#include <zvec/ailego/container/params.h>
 #include <zvec/ailego/logger/logger.h>
-#include <zvec/core/interface/index.h>
-#include <zvec/db/index_params.h>
 #include <zvec/db/status.h>
 #include <zvec/db/type.h>
-#if RABITQ_SUPPORTED
-#include "core/algorithm/hnsw_rabitq/rabitq_params.h"
-#endif
 #include "db/common/constants.h"
 #include "db/common/file_helper.h"
 #include "db/common/global_resource.h"
 #include "db/common/typedef.h"
 #include "db/index/column/inverted_column/inverted_indexer.h"
-#include "db/index/column/vector_column/engine_helper.hpp"
 #include "db/index/column/vector_column/vector_column_indexer.h"
 #include "db/index/common/index_filter.h"
 #include "db/index/common/meta.h"
 #include "db/index/storage/forward_writer.h"
-#include "zvec/core/framework/index_factory.h"
-#include "zvec/core/framework/index_meta.h"
-#include "zvec/core/framework/index_provider.h"
-#include "zvec/core/framework/index_reformer.h"
 #include "roaring.hh"
 
 namespace zvec {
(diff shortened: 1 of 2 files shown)


iaojnh (approved):
LGTM

Issue #426, opened 2026-05-27
[BUG]: query() returns empty fields for columns added via add_column()


feihongxu0824:
## Bug Description

After adding a field via `Collection.add_column()`, inserting documents with values for that field, and then querying, the `query()` results return empty `fields: {}` — the dynamically added column's values are missing. However, `fetch()` returns the field values correctly.

If the same field is defined at schema creation time (via `CollectionSchema`), both `query()` and `fetch()` return the values correctly.

## Environment

- zvec version: 0.4.0
- Python: 3.14.2
- OS: macOS (arm64)

## Reproduction

```python
import os
import tempfile
import zvec

# --- Case 1: add_column after creation (BUG) ---
path = os.path.join(tempfile.mkdtemp(), "collection")
schema = zvec.CollectionSchema(
    name="test",
    vectors=zvec.VectorSchema("embedding", zvec.DataType.VECTOR_FP32, 4),
)
col = zvec.create_and_open(path=path, schema=schema)

col.add_column(zvec.FieldSchema("score", zvec.DataType.INT64, nullable=True))

col.insert([
    zvec.Doc(id="doc_1", fields={"score": 100}, vectors={"embedding": [0.1, 0.2, 0.3, 0.4]}),
    zvec.Doc(id="doc_2", fields={"score": 200}, vectors={"embedding": [0.5, 0.6, 0.7, 0.8]}),
])

# fetch works correctly
print(col.fetch("doc_1"))
# -> fields: {"score": 100} ✅

# query returns empty fields
results = col.query(
    zvec.VectorQuery("embedding", vector=[0.1, 0.2, 0.3, 0.4]),
    topk=2,
    output_fields=["score"],
)
for r in results:
    print(r)
# -> fields: {} ❌

col.destroy()

# --- Case 2: field defined at schema creation (WORKS) ---
path2 = os.path.join(tempfile.mkdtemp(), "collection")
schema2 = zvec.CollectionSchema(
    name="test2",
    fields=zvec.FieldSchema("score", zvec.DataType.INT64, nullable=True),
    vectors=zvec.VectorSchema("embedding", zvec.DataType.VECTOR_FP32, 4),
)
col2 = zvec.create_and_open(path=path2, schema=schema2)

col2.insert([
    zvec.Doc(id="doc_1", fields={"score": 100}, vectors={"embedding": [0.1, 0.2, 0.3, 0.4]}),
[...]

PR #429, opened 2026-05-28
fix: sync querier schema after column DDL to fix empty query fields

chinaux:
fix: (#426)

After add_column/drop_column/alter_column, the querier's cached schema
was not updated, causing query() to return empty fields while fetch()
  worked correctly. Add schema sync and corresponding test coverage.

### DIFF
--- a/python/zvec/model/collection.py
+++ b/python/zvec/model/collection.py
@@ -168,6 +168,7 @@ class Collection:
         """
         self._obj.AddColumn(field_schema._get_object(), expression, option)
         self._schema = CollectionSchema._from_core(self._obj.Schema())
+        self._querier._schema = self._schema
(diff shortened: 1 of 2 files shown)


python/zvec/model/collection.py
zhourrr, line 132: create index 和 drop index 是否也需要更新 querier schema

chinaux: querier._schema 被用于 convert_to_py_doc(doc, schema)，该函数遍历 schema.forward_fields() 提取字段值——关心的是字段名和数据类型，和索引无关。为防止未来 schema 变更逻辑， 还是加上吧

zhourrr (approved):

PR #432, opened 2026-05-31
refactor: clarify segment-local row ID handling and tests and fixes bugs

### LINKED ISSUE #430: [BUG]: Sparse vector browse fails with Can't find block for doc_id after repeated inserts
### What happened?

After creating a sparse-vector collection, optimizing it, and then inserting two more small batches, a browse/list-style query that requests vector payloads fails with:

```text
RuntimeError: Read record batch failed: ExecutionError in Gandiva: read next batch failed:ExecutionError in Gandiva: fetch vector failed:Can't find block for doc_id:16
```

The same query succeeds after the first two-doc insert, and fails after the second two-doc insert.

### Minimal reproduction

This script only depends on `zvec` and Python standard library modules. It creates a fresh collection path from scratch.

```python
from __future__ import annotations

from pathlib import Path
import sys
import tempfile

import zvec


def make_doc(i: int) -> zvec.Doc:
    return zvec.Doc(
        id=str(i),
        vectors={
            "embedding": {
                i % 768: 0.1 + i * 0.001,
                (i + 100) % 768: 0.2,
                (i + 200) % 768: 0.3,
            }
        },
    )


def browse(collection: zvec.Collection, label: str) -> None:
    print(label)
    rows = collection.query(
        vectors=None,
        topk=200,
        filter=None,
        include_vector=True,
        output_fields=None,
    )
    print("browse rows:", len(rows))
    print("browse ids:", [row.id for row in rows])


def fresh_path() -> Path:
    base = Path(tempfile.gettempdir())
    for i in range(1, 1000):
        path = base / f"zvec_gandiva_missing_block_repro_{i:03d}"
        if not path.exists():
            return path
    raise RuntimeError("Could not find a fresh repro path.")


def main() -> None:
    path = Path(sys.argv[1]) if len(sys.argv) > 1 else fresh_path()
    if path.exists():
        raise SystemExit(f"Refusing to reuse existing path: {path}")

        "fresh_sparse_repro",
        vectors=zvec.VectorSchema(
            "embedding",
            zvec.DataType.SPARSE_VECTOR_FP32,
            768,
            zvec.HnswIndexParam(
                metric_type=zvec.MetricType.IP,
                m=50,
                ef_construction=500,
            ),
        ),
    )

    collection = zvec.create_and_open(str(path), schema)
    print("path:", path)
    print("schema:", collection.schema)

    initial_docs = [make_doc(i) for i in range(1, 15)]
    print("insert initial:", [doc.id for doc in initial_docs])
    print("insert statuses:", collection.insert(initial_docs))
    collection.flush()
    print("stats after initial flush:", collection.stats)

    collection.optimize(zvec.OptimizeOption())
    print("stats after optimize:", collection.stats)

    first_pair = [make_doc(15), make_doc(16)]
    print("insert first pair:", [doc.id for doc in first_pair])
    print("insert statuses:", collection.insert(first_pair))
    collection.flush()
    print("stats after first pair:", collection.stats)
    browse(collection, "browse after first pair:")

    second_pair = [make_doc(17), make_doc(18)]
    print("insert second pair:", [doc.id for doc in second_pair])
    print("insert statuses:", collection.insert(second_pair))
    collection.flush()
    print("stats after second pair:", collection.stats)

    browse(collection, "browse after second pair:")


if __name__ == "__main__":
    main()
```

### Output

```text
stats after optimize: {"doc_count":14, "index_completeness":{"embedding":1.000000}}
stats after first pair: {"doc_count":16, "index_completeness":{"embedding":0.875000}}
browse after first pair:
browse rows: 16
stats after second pair: {"doc_count":18, "index_completeness":{"embedding":0.777778}}
browse after second pair:
RuntimeError: Read record batch failed: ExecutionError in Gandiva: read next batch failed:ExecutionError in Gandiva: fetch vector failed:Can't find block for doc_id:16
```

### Expected behavior

`collection.query(vectors=None, topk=200, filter=None, include_vector=True, output_fields=None)` should return all matching docs with vector payloads, or at least not fail with a missing block for a valid doc id.

### Environment

- `zvec` Python package: `0.4.0`
- Python: `3.10`
- OS: macOS

### Notes

The failure is specific to reading vector payloads. A query with `include_vector=False` succeeds in the same state.

zhourrr:
Fix row-id translation across combined segment blocks. Combined scans now emit segment-local LOCAL_ROW_ID values, preserving correctness for include-vector fetch and delete filtering. Also translate per-block group-by callbacks to segment-local ids and add regression coverage for multi-block scans with nonzero document ids.
fixes #430

### DIFF
--- a/src/db/common/constants.h
+++ b/src/db/common/constants.h
@@ -26,6 +26,8 @@ const uint32_t MIN_MEMORY_LIMIT_BYTES = 100 * 1024 * 1024;
 
 const uint64_t INVALID_DOC_ID = UINT64_MAX;
 
+// Synthetic local row id column exposed as "_zvec_row_id_". Higher layers
+// define which local ID space it represents at their boundary.
 const std::string LOCAL_ROW_ID = "_zvec_row_id_";
 
 const std::string GLOBAL_DOC_ID = "_zvec_g_doc_id_";
--- a/src/db/common/constants.h
+++ b/src/db/common/constants.h
@@ -26,8 +26,6 @@ const uint32_t MIN_MEMORY_LIMIT_BYTES = 100 * 1024 * 1024;
 
 const uint64_t INVALID_DOC_ID = UINT64_MAX;
 
-// Synthetic local row id column exposed as "_zvec_row_id_". Higher layers
-// define which local ID space it represents at their boundary.
 const std::string LOCAL_ROW_ID = "_zvec_row_id_";
 
 const std::string GLOBAL_DOC_ID = "_zvec_g_doc_id_";
--- a/src/db/index/column/vector_column/combined_vector_column_indexer.cc
+++ b/src/db/index/column/vector_column/combined_vector_column_indexer.cc
@@ -116,6 +116,18 @@ Result<IndexResults::Ptr> CombinedVectorColumnIndexer::Search(
       }
     }
 
+    std::unique_ptr<vector_column_params::GroupByParams> group_by;
+    if (query_params.group_by) {
+      auto group_by_func = query_params.group_by->group_by;
+      auto block_offset = block_offsets_[i];
+      group_by = std::make_unique<vector_column_params::GroupByParams>(
+          query_params.group_by->group_topk, query_params.group_by->group_count,
+          [group_by_func = std::move(group_by_func),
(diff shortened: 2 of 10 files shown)


chinaux (approved):
LGTM

PR #441, opened 2026-06-01
perf(fts): add block-max skip and score early-exit to ConjunctionIterator

egolearner:
- Block-max skip: skip entire non-competitive blocks (128 docs) in do_next() by checking block_max_info_for() score upper bounds
- Score early-exit: short-circuit score() accumulation when remaining upper bound cannot beat the threshold
- Phrase forwarding: PhraseDocIterator propagates min_competitive_score to its inner conjunction, enabling the above optimizations for phrase queries
- optIsRequired: when must-only block_max < threshold, promote should clauses to required so docs without any should match are skipped
- DisjunctionIterator::advance() bypasses the WAND loop to avoid incorrectly pruning target docs when used as a should clause

Benchmark (500k docs, quora dataset, BitPacked mode):
  AND queries: 22-38% faster (topk=10: 3.65s→2.86s, topk=3: 3.63s→2.26s)
  Phrase queries: 33% faster (205s→137s)

### DIFF
--- a/src/db/index/column/fts_column/iterator/fts_conjunction_iterator.cc
+++ b/src/db/index/column/fts_column/iterator/fts_conjunction_iterator.cc
@@ -55,7 +55,7 @@ uint32_t ConjunctionIterator::next_doc() {
 
   // Advance the lead iterator and try to find agreement
   uint32_t candidate = must_iterators_[0]->next_doc();
-  cached_doc_id_ = do_next(candidate);
+  cached_doc_id_ = do_next(candidate, true);
   return cached_doc_id_;
 }
 
@@ -75,7 +75,7 @@ uint32_t ConjunctionIterator::next_doc(const zvec::IndexFilter *filter) {
   // reach do_next() alignment.
   uint32_t candidate = must_iterators_[0]->next_doc(filter);
   while (candidate != NO_MORE_DOCS) {
-    candidate = do_next(candidate);
+    candidate = do_next(candidate, true);
     if (candidate == NO_MORE_DOCS || !filter->is_filtered(candidate)) {
       break;
     }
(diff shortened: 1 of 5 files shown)


src/db/index/column/fts_column/iterator/fts_disjunction_iterator.cc
```diff
@@ -223,7 +223,29 @@ uint32_t DisjunctionIterator::advance(uint32_t target) {
       iter->advance(target);
     }
   }
-  return next_doc();
+
+  // Bypass the WAND loop — advance() is a seek operation that must not
+  // apply competitive-score pruning.  When this DisjunctionIterator serves
+  // as a should clause, the caller (ConjunctionIterator::score()) relies on
+  // advance() returning target if it exists; WAND block-max pruning would
+  // incorrectly skip target and silently lose the should score.
```
JalinWang, line 231: Should ConjunctionIterator::advance() also maintain the seek-only semantics?

src/db/index/column/fts_column/iterator/fts_conjunction_iterator.cc
```diff
@@ -104,6 +104,59 @@ uint32_t ConjunctionIterator::advance(uint32_t target) {
...
+      must_block_max_sum_ = must_only_sum;
+      opt_is_required_ = (!should_iterators_.empty() &&
+                          must_block_max_sum_ < min_competitive_score_);
+      return candidate;
+    }
+
+    // Current block is non-competitive, skip to the next block
+    uint32_t next_block_start = min_block_end + 1;
+    if (next_block_start == 0) {
+      // overflow: min_block_end was MAX → 已耗尽
```
JalinWang: Chinese characters

src/db/index/column/fts_column/iterator/fts_disjunction_iterator.cc
egolearner: fixed

src/db/index/column/fts_column/iterator/fts_conjunction_iterator.cc
egolearner: fixed

JalinWang (approved):

PR #456, opened 2026-06-03
fix(segment): use raw vectors for RaBitQ merge during compaction

JalinWang:
## Summary
- During segment compaction, `ReduceVectorIndex` was feeding already-encoded RaBitQ data back through the HNSW_RABITQ target indexer's reformer, which re-quantized it and produced garbage — resulting in recall = 0.
- Fix: when the quantize type is RABITQ, source merge data from the flat (raw fp32) vector indexers instead of the quant vector indexers, so the target's internal reformer receives the original vectors it expects.
- Re-enables the previously disabled `SegmentCompactReuseTest` for HNSW_RABITQ.

### DIFF
--- a/src/db/index/segment/segment_helper.cc
+++ b/src/db/index/segment/segment_helper.cc
@@ -716,12 +716,9 @@ Status SegmentHelper::ReduceVectorIndex(
       auto vector_quan_index_path = FileHelper::MakeQuantizeVectorIndexPath(
           output_segment_path, field->name(), vector_quan_block_id);
 
-      // For RABITQ: the target HNSW_RABITQ index's add_with_id_impl expects
-      // raw fp32 input (its internal reformer handles the quantization).
-      // Using get_quant_vector_indexer would return existing HNSW_RABITQ
-      // indexers whose stored data is already rabitq-encoded; feeding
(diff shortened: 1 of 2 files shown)


src/db/index/segment/segment_helper.cc
```diff
@@ -716,10 +716,20 @@ Status SegmentHelper::ReduceVectorIndex(
       auto vector_quan_index_path = FileHelper::MakeQuantizeVectorIndexPath(
           output_segment_path, field->name(), vector_quan_block_id);
 
-      s = MergeWithOptionalReuse(
-          vector_quan_index_path, *field_for_quantize,
-          collect_merge_indexers(&Segment::get_quant_vector_indexer), filter,
-          concurrency, nullptr);
+      // For RABITQ: the target HNSW_RABITQ index's add_with_id_impl expects
```
egolearner: 可以简化一下注释，类似"rabitq需要原始fp32向量作为输入，其他类型需要量化后的向量作为输入"

JalinWang: done

egolearner (approved):
LGTM

PR #478, opened 2026-06-09
feat(query): add VectorViewClause zero-copy path and unify validate

egolearner:
- Add VectorViewClause (string_view-based) as zero-copy counterpart to VectorClause; variant now holds VectorClause | VectorViewClause | FtsClause
- Add QueryTarget::get_vector_view() unified accessor via std::visit, returns optional<VectorViewClause> regardless of which variant is held
- Split validate_and_sanitize into QueryTarget::validate (read-only) + sanitize_sparse_vector (mutate); validate handles both VectorClause and VectorViewClause via get_vector_view()
- Collection::Query passes original request directly to sqlengine when no sparse sanitization is needed; only copies when sort is required
- Change build_query_info/BuildSQLInfoFromSearchQuery to take const SearchQuery& so VectorMatrixNode string_views point to caller's data
- sqlengine internals use get_vector_view() instead of get_vector_view_clause()

### DIFF
--- a/src/db/collection.cc
+++ b/src/db/collection.cc
@@ -1704,6 +1704,11 @@ Result<DocPtrList> CollectionImpl::Query(const MultiQuery &query) const {
         query.queries.size()));
   }
 
+  if (auto s = validate_topk_and_output_fields(query.topk, query.output_fields);
+      !s.ok()) {
+    return tl::make_unexpected(s);
+  }
+
   auto segments = get_all_segments();
   if (segments.empty()) {
     return DocPtrList();
--- a/src/db/collection.cc
+++ b/src/db/collection.cc
@@ -1667,14 +1667,14 @@ Result<DocPtrList> CollectionImpl::Query(const SearchQuery &query) const {
 
   CHECK_DESTROY_RETURN_STATUS_EXPECTED(destroyed_, false);
 
-  SearchQuery sanitized = query;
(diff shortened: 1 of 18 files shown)


src/db/collection.cc
```diff
@@ -1724,8 +1736,10 @@ Result<DocPtrList> CollectionImpl::Query(const MultiQuery &query) const {
     sq.include_doc_id_ = query.include_doc_id_;
     sq.output_fields_ = query.output_fields;
```
feihongxu0824, line 1742: 这里是不是漏了对SearchQuery的validate？

src/db/index/common/type_helper.cc
```diff
@@ -26,19 +26,6 @@ bool sort_and_find_duplicates(uint32_t *indices, char *values, size_t n,
   if (n <= 1) {
     return false;
   }
-  bool already_sorted = true;
```
feihongxu0824, line 29: 写入过程应该还需要这段逻辑

egolearner: 改为先调用need_sanitize_sparse判断是否需要排序

src/db/collection.cc
egolearner: fixed

feihongxu0824 (approved):

PR #479, opened 2026-06-09
refactor: disable diskann index create on open

richyreachy:
disable diskann index create on open


python/zvec/model/collection.py
```diff
@@ -128,6 +129,9 @@ def create_index(
                 Defaults to ``IndexOption()``.
 
         """
+        # Reject DiskAnn indexes on unsupported platforms (e.g. ARM) at
```
feihongxu0824: 判断是否支持diskann的逻辑不应该放在python层，需要放在c++层，不然所有sdk都得重复写这个逻辑

richyreachy: changed to c++ implementation.

richyreachy: update to test if aio exists

zhourrr (approved):

PR #489, opened 2026-06-10
fix(c_api): use IndexParams::clone() to simplify convert_c_index_params_to_cpp

egolearner:
convert_c_index_params_to_cpp used a whitelist switch over IndexType, prone to miss for newly index types.

Replace the entire switch with a single clone() call (pure-virtual on IndexParams base, implemented by all subclasses).


chinaux (approved):
lgtm

PR #496, opened 2026-06-15
fix(hnsw): restore persisted info on reopen for UniformInt8 quantizer

luoxiaojian:


tests/db/collection_test.cc
```diff
@@ -2803,6 +2811,12 @@ TEST_F(CollectionTest, Feature_Optimize_Repeated) {
         enable_mmap, std::make_shared<HnswRabitqIndexParams>(MetricType::IP, 7,
                                                              256, 16, 200, 0));
 #endif
+    // Vamana + UniformInt8: VamanaStreamer::open already restores reformer
```
JalinWang: the same

tests/db/collection_test.cc
```diff
@@ -2787,6 +2787,14 @@ TEST_F(CollectionTest, Feature_Optimize_Repeated) {
     run_repeated_optimize_test(
         enable_mmap, std::make_shared<HnswIndexParams>(MetricType::IP, 16, 200,
                                                        QuantizeType::FP16));
+    // Regression: UniformInt8 quantizer. Its reformer params (scale/bias) are
```
JalinWang: It seems that DB layer currently isn't aware about UniformInt8 🤔 Should it be ordinary Int8 quantization here?

JalinWang (approved):
LGTM

PR #498, opened 2026-06-15
fix(db): guard against incompatible query param type in vector search

egolearner:
Add null-check after dynamic_cast in each vector index branch (HNSW, HNSW_RABITQ, IVF, DiskANN, Vamana). Previously a mismatched query param type would cause a nullptr dereference; now it returns InvalidArgument with a descriptive message.


feihongxu0824 (approved):

egolearner:
close in favor of #518

PR #502, opened 2026-06-17
fix(python): validate numpy vector queries safely

HosniBelfeki:
## Summary

This is a small Python SDK validation hardening change.

When `Query._validate()` receives both `id` and a numpy vector, the previous truthiness check (`self.id and self.vector`) can trigger numpy's ambiguous truth-value error before Zvec raises its intended validation error.

This switches the check to the existing `has_id()` / `has_vector()` helpers and adds a regression test.

## Tests

- `python -m ruff check python\zvec\model\param\query.py python\tests\test_params.py`
- `$env:PYTHONPATH='python'; python -m pytest -o addopts= python/tests/test_params.py::TestQuery python/tests/test_fts_query.py::TestFtsQueryValidation -q`

### DIFF
--- a/python/tests/test_params.py
+++ b/python/tests/test_params.py
@@ -428,6 +428,11 @@ class TestQuery:
         with pytest.raises(ValueError):
             vq._validate()
 
+    def test_validate_fails_on_both_id_and_numpy_vector(self):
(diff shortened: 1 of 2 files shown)


Cuiyus (approved):
LGTM

PR #511, opened 2026-06-22
fix(python): install the _zvec extension inside the zvec package

Cuiyus:
## Summary

The compiled extension `_zvec*.so` was installed at the **site-packages root** (and the DiskANN plugin next to it), polluting the top-level namespace. It was imported everywhere as the bare top-level module `_zvec`.

This moves the extension **into the package** (`zvec/_zvec*.so`) and imports it as `zvec._zvec`.

## Changes

- `CMakeLists.txt`: install `_zvec` (and the DiskANN plugin) into `${ZVEC_PY_INSTALL_DIR}/zvec` instead of the platlib root. The plugin stays next to `_zvec.so`, so the existing `$ORIGIN` rpath (`src/binding/python/CMakeLists.txt`) still resolves it at runtime.
- Import sites updated (19 files total):
  - 16 `.py` files: `from _zvec[...] import` → `from zvec._zvec[...] import`
  - 2 `.pyi` stubs: `import _zvec.*` and qualified annotations (`_zvec.typing.DataType`, …) → `zvec._zvec.*`

pybind11 3.0 names its `def_submodule`s by the actual import path, so `from zvec._zvec.param import _SearchQuery` resolves correctly — confirmed empirically before the rewrite.

Docstrings / error strings that merely mention `_zvec` were intentionally left as prose.

## Verification (macOS arm64)

- ✅ Wheel places the extension at `zvec/_zvec.cpython-311-darwin.so`; **nothing** left at the site-packages root
- ✅ `import zvec`, `import zvec._zvec`, `from zvec._zvec.param import _SearchQuery`, and the public API all import
- ✅ Full Python test suite: **1092 passed, 70 skipped** (skips are libaio/DiskANN-only cases, expected on macOS)

> Note: this branch is cut from `main` and is independent of the two sibling packaging PRs (exclude SDK `include/`+`lib/`; gate the DiskANN stub). They touch the same install block and will need a trivial rebase depending on merge order.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,7 +174,10 @@ if(BUILD_PYTHON_BINDINGS)
     # COMPONENT python: only these runtime artifacts are pulled into the wheel
     # (see install.components in pyproject.toml). The cc_library(PACKED) SDK
     # install rules use the default component and are excluded from the wheel.
-    install(TARGETS _zvec LIBRARY DESTINATION ${ZVEC_PY_INSTALL_DIR}
+    # Install the extension inside the zvec package (zvec/_zvec*.so) rather than
+    # at the site-packages root, so it does not pollute the top-level namespace.
+    # The Python code imports it as `zvec._zvec` accordingly.
+    install(TARGETS _zvec LIBRARY DESTINATION ${ZVEC_PY_INSTALL_DIR}/zvec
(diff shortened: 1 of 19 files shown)


egolearner (approved):
LGTM

PR #519, opened 2026-06-23
fix(python): handle missing query-by-id documents

HosniBelfeki:
## Summary

Handle missing documents in Python query-by-id execution with the intended `ValueError`.

Previously, when `collection.Fetch([query.id])` returned no document, `_build_search_query()` called `next(iter(fetched.values()))` and raised a raw `StopIteration` before reaching the existing user-facing error message.

This uses `next(..., None)` so missing IDs raise:

`ValueError: Document with id '<id>' not found`

A focused regression test was added for the missing-id path.

## Tests

- `python -m ruff check python\zvec\executor\query_executor.py python\tests\test_query_executor.py`
- `$env:PYTHONPATH='python'; python -m pytest -o addopts= python/tests/test_query_executor.py -q`

### DIFF
--- a/python/tests/test_query_executor.py
+++ b/python/tests/test_query_executor.py
@@ -300,3 +300,16 @@ class TestQueryExecutor:
             results = executor._execute_python_pipeline(vectors, collection)
         assert results == [["raw1"], ["raw2"]]
         assert collection.Query.call_count == 2
+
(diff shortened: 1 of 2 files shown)


egolearner (approved):
LGTM

PR #523, opened 2026-06-24
chore(submodule): ignore dirty submodule worktrees

JalinWang:
Set `ignore = dirty` for every entry in `.gitmodules` so local submodule modifications do not show up as repository dirtiness.

This keeps the submodule status noise down while leaving tracked changes intact.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,54 +1,64 @@
 [submodule "thirdparty/googletest/googletest-1.10.0"]
 	path = thirdparty/googletest/googletest-1.10.0
 	url = https://github.com/google/googletest.git
+	ignore = dirty
 [submodule "thirdparty/sparsehash/sparsehash-2.0.4"]
 	path = thirdparty/sparsehash/sparsehash-2.0.4
 	url = https://github.com/sparsehash/sparsehash.git
-	ignore = untracked
+	ignore = dirty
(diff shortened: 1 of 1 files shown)


egolearner:
Thanks for the PR!

I don’t think we should set `ignore = dirty` for all thirdparty submodules by default. In this repo, letting `git status` report dirty submodule worktrees is intentional: it helps us catch accidental or missed local changes under `thirdparty/` before they slip through review/release work. So the current behavior is more of a safeguard/feature than a bug.

Could you share the concrete scenario where these submodules need to be fully ignored? If this is about local build artifacts or tool-generated noise in a specific submodule, we can consider a narrower fix instead of changing the repo-wide default.

JalinWang:
> Could you share the concrete scenario where these submodules need to be fully ignored? If this is about local build artifacts or tool-generated noise in a specific submodule, we can consider a narrower fix instead of changing the repo-wide default.

`sparsehash` and `googletest` can end up with local changes on Windows because of `apply_patch`, which keeps `git status` dirty. `ignore = dirty` is the right fit here.

I also think `ignore = all` is too broad for these cases, because submodule commit changes are something we still want to track.

Updated this PR to be limited to the submodules that receive patches.

egolearner (approved):

PR #530, opened 2026-06-25
fix(diskann, hnsw sparse builder): lost-wakeup race in builder progress loops

JalinWang:
## Problem

DiskAnn / HNSW-sparse / PQ-cluster builds stall for 15–30 s on small datasets due to a lost-wakeup race in the condition-variable progress loops.

Two bugs:

1. `multi_chunk_cluster.h` workers never call `cond_.notify_one()` after `(*finished)++`.
2. All loops use `task_group->is_finished()` as the wait predicate — this races because the task-group internal `notify()` fires *after* the closure returns, so a wakeup triggered inside the closure is missed on re-check.

## Fix

1. Add `{ lock_guard lk(mutex_); cond_.notify_one(); }` after each `(*finished)++` in cluster workers.
2. Replace `while (!task_group->is_finished())` with `while (finished.load() < count)` inside the locked scope across all affected builders.

## Files changed

| File | Change |
|------|--------|
| `multi_chunk_cluster.h` | Add `notify_one()` in 4 template workers |
| `multi_chunk_cluster.cc` | Fix predicate in `cluster()` and `label()` |
| `diskann_builder.cc` | Fix predicate in `build_internal()` and `prune_internal()` |
| `hnsw_sparse_builder.cc` | Fix predicate in `build_graph()` |
| `diskann_builder_test.cc` | Regression test: 12-vector build must finish < 5 s |

## Result

`DiskAnnBuilderTest.SmallDatasetBuildTime`: 30 s → 82 ms

### DIFF
--- a/src/core/algorithm/cluster/multi_chunk_cluster.cc
+++ b/src/core/algorithm/cluster/multi_chunk_cluster.cc
@@ -201,16 +201,18 @@ int MultiChunkClusterAlgorithm::cluster(IndexThreads::Pointer threads,
                              threads->count(), &cents, &finished));
   }
 
-  while (!task_group->is_finished()) {
+  {
     std::unique_lock<std::mutex> lk(mutex_);
-    cond_.wait_until(lk, std::chrono::system_clock::now() +
-                             std::chrono::seconds(check_interval_secs_));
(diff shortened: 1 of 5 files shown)


richyreachy (approved):
LGTM

PR #531, opened 2026-06-25
fix(sql engine): misused group_count/group_topk

JalinWang:
`group_count` represents the number of groups to return, not the number of documents per group, while `group_topk` represents the number of documents per group.


egolearner (approved):
LGTM

PR #538, opened 2026-06-29
fix(python): validate queries during execution

HosniBelfeki:
## Summary

Run `Query._validate()` in the Python query execution path before building the core `_SearchQuery`.

`Query` already defines validation rules such as rejecting queries that provide both `id` and `vector`, but `_build_search_query()` did not call `_validate()`. As a result, invalid query objects could continue into execution instead of failing early with the intended Python `ValueError`.

This validates each query before conversion and adds a focused regression test.

## Tests

- `python -m ruff check python\zvec\executor\query_executor.py python\tests\test_query_executor.py`
- `python -m py_compile python\zvec\executor\query_executor.py python\tests\test_query_executor.py`

Note: full pytest requires a locally built `zvec._zvec` extension matching current `main`; the installed wheel is older than the current source layout.

### DIFF
--- a/python/tests/test_query_executor.py
+++ b/python/tests/test_query_executor.py
@@ -313,3 +313,17 @@ class TestQueryExecutor:
             executor._build_search_query(
                 ctx, Query(field_name="test", id="missing"), collection
             )
+
+    def test_build_search_query_validates_query(self):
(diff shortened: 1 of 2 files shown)


Cuiyus (approved):
LGTM

PR #539, opened 2026-06-29
feat(tools): opt-in SquaredEuclidean micro-benchmark (tools/core)

### LINKED ISSUE #535: Proposal: opt-in benchmarks/ harness for per-query scan throughput (ThreadPool + SIMD dispatch)
## Proposal: opt-in `benchmarks/` harness for zvec's own primitives

Before opening a PR I'd like to check whether you'd welcome a small,
**opt-in** benchmark harness under a new top-level `benchmarks/` directory. I
have a branch ready and would only send it if this is something you want
upstream — happy to drop it otherwise.

### What it is

A single micro-benchmark, `distance_scan_bench`, that measures flat-scan
throughput **the way zvec actually scales it** — and nothing more:

- parallelism is applied **per query** across `ailego::ThreadPool`;
- each individual query is a **serial** scan that calls zvec's
  **SIMD-dispatched** `Distance::SquaredEuclidean` (runtime `CpuFeatures`
  dispatch: SSE / AVX / AVX512 / NEON).

No OpenMP, no hand-written intrinsics, no kernel changes. It uses exactly the
two primitives zvec already relies on. This is deliberately aligned with the
architecture (`ENABLE_OPENMP` off, hot kernels hand-vectorized, throughput from
distributing *queries* over the pool while the single-query scan stays serial)
rather than cutting across it.

```mermaid
flowchart TD
    subgraph POOL["ailego::ThreadPool — parallel ACROSS queries"]
      direction LR
      q0["query 0<br/>serial scan"]:::par
      q1["query 1<br/>serial scan"]:::par
      qn["query N-1<br/>serial scan"]:::par
    end
    POOL --> K
    subgraph K["each scan: one query vs all DB vectors"]
      direction TB
      loop["for i in 0..num_db&nbsp;&nbsp;(serial)"]:::seq
      kern["Distance::SquaredEuclidean(db_i, q, dim)<br/>SIMD-dispatched: SSE / AVX / AVX512 / NEON"]:::simd
      loop --> kern
    end
    classDef par  fill:#1f7a1f,color:#fff,stroke:#0a0,stroke-width:2px
    classDef seq  fill:#2b2b2b,color:#fff,stroke:#888
    classDef simd fill:#1f3a7a,color:#fff,stroke:#46f,stroke-width:2px
```

The pool parallelizes the outer axis (queries); the SIMD kernel vectorizes the
inner axis (one pair's reduction). They compose and neither touches the other's
loop, so the parallel result is **bit-identical** to serial — each query writes a
disjoint output row and the per-pair reduction order is unchanged. The benchmark
self-validates this on every run (and is registered as a `ctest`), exiting
non-zero on any mismatch.

### Zero impact when off

- New CMake option `BUILD_BENCHMARKS`, **default `OFF`** — the target does not
  exist unless explicitly enabled, so the default build and CI matrix are
  untouched.
- Links only `zvec_ailego`; adds no third-party dependency.

```bash
cmake -S . -B build -DBUILD_BENCHMARKS=ON -DCMAKE_BUILD_TYPE=Release
cmake --build build --target distance_scan_bench
./build/bin/distance_scan_bench            # default 4096 256 128
ctest -R distance_scan_bench               # self-validation
```

### Measured (illustrative)

`num_db=8192, num_query=256, dim=128`, Apple Silicon (10 cores), kernel SIMD
path **NEON**, `-O3`, best-of-5; serial baseline `0.023 s`:

| Threads | Speedup | max_abs_diff vs serial |
|--------:|--------:|-----------------------:|
| 1 | 1.00× | 0.000e+00 |
| 2 | 1.91× | 0.000e+00 |
| 4 | 3.76× | 0.000e+00 |
| 8 | 4.31× | 0.000e+00 |

```mermaid
xychart-beta
    title "Per-query thread-pool scaling (NEON) — bar = measured, line = ideal"
    x-axis "Threads" [1, 2, 4, 8]
    y-axis "Speedup x" 0 --> 8
    bar [1.00, 1.91, 3.76, 4.31]
    line [1, 2, 4, 8]
```

Near-linear to the physical-core count, then flattens as the scan becomes
memory-bandwidth-bound. Inputs are synthetic and deterministic (no RNG);
correctness is established by bit-for-bit comparison against the serial
reference, not a data cohort.

### Questions for maintainers

1. Is a top-level `benchmarks/` directory welcome, or would you prefer a
   different location / naming?
2. Is the opt-in `BUILD_BENCHMARKS=OFF` model acceptable, or do you have an
   existing convention for performance tooling I should follow?
3. Any scope preferences (e.g. keep it to this one scan, or none at all)?

If this is of interest I'll open the PR. This follows the same
architecture-alignment approach as the rotator fix in #534.

---

*The "parallelize queries, never the reduction" schedule was derived and proven
legal (no carried dependence) with a polyhedral auto-scheduling pass —
[cluster_compilot](https://github.com/cluster2600/cluster_compilot), an
implementation of Agentic Auto-Scheduling (arXiv:2511.00592). The benchmark
implements that schedule with zvec's existing `ThreadPool` + SIMD-dispatched
kernels; it proposes no kernel change.*

cluster2600:
Implements the opt-in micro-benchmark proposed in #535, following the placement and scope decisions from the issue thread.

## What

`tools/core/distance_bench` — a standalone tool that models a brute-force scan and times the SIMD-dispatched `zvec::ailego::Distance::SquaredEuclidean` against a plain scalar reference across a range of dims, **self-validating that the SIMD result matches the scalar one** before reporting any timing.

## Decisions from #535

- **Location: `tools/core/`** — the benchmarked kernel (`Distance::SquaredEuclidean`) lives in `src/ailego/math/distance.h`, so this is a core-layer micro-benchmark (not the db layer).
- **Build flag: reuses `BUILD_TOOLS`** — added as a `cc_binary` under `tools/core/`; no new flag.
- **Not a `ctest`** — micro-benchmark numbers are machine dependent, so it stays a standalone executable. It exits non-zero only when the SIMD kernel disagrees with the scalar reference (`--tolerance`, default `1e-4` relative).

## Sample run (Apple Silicon / NEON, `--num_db=8192 --reps=50`)

```
   dim   scalar(ms)     simd(ms)   speedup
    16        1.594        1.554     1.03x
    31        3.499        1.613     2.17x
    64        8.923        2.094     4.26x
   128       26.486        4.755     5.57x
   257       59.054       11.504     5.13x
```

The SIMD win grows with dim as expected (and is larger on AVX-512 x86). Small dims sit near 1x where dispatch overhead dominates — surfacing exactly the kind of crossover this benchmark is meant to make visible.

## Usage

```
./distance_bench --num_db=8192 --dims=1,2,3,7,16,31,64,128,257 --reps=50
```

Closes #535

### DIFF
--- a/tools/core/CMakeLists.txt
+++ b/tools/core/CMakeLists.txt
@@ -10,6 +10,14 @@ cc_binary(
   )
 
 cc_binary(
+    NAME distance_bench
+    STRICT PACKED
+    SRCS distance_bench.cc
+    INCS ${PROJECT_ROOT_DIR}/src/
+    LIBS gflags zvec_ailego
+  )
+
+cc_binary(
     NAME local_builder
     STRICT PACKED
     SRCS local_builder.cc
--- a/tools/core/distance_bench.cc
+++ b/tools/core/distance_bench.cc
@@ -1,0 +1,133 @@
+// Copyright 2025-present the zvec project
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Opt-in micro-benchmark for the core FP32 SquaredEuclidean distance kernel.
+//
+// Models a brute-force scan: for each query, the squared-Euclidean distance to
+// every database vector is computed via
+// zvec::ailego::Distance::SquaredEuclidean (which dispatches to AVX / AVX-512 /
+// NEON). We report its throughput against a plain scalar reference across a
+// range of dims, and self-validate that the SIMD result agrees with the scalar
+// one before trusting any timing.
+//
+// Standalone executable (built under BUILD_TOOLS); intentionally NOT registered
+// as a ctest -- micro-benchmark numbers are machine dependent. It exits
+// non-zero only if the SIMD kernel disagrees with the scalar reference.
+//
+//   ./distance_bench --num_db=8192 --dims=1,2,3,7,16,31,64,128,257 --reps=50
+
+#include <algorithm>
+#include <chrono>
+#include <cmath>
+#include <cstddef>
+#include <cstdio>
+#include <random>
+#include <sstream>
+#include <string>
+#include <vector>
+#include <ailego/math/distance.h>
+#include <gflags/gflags.h>
+
+DEFINE_int32(num_db, 8192, "database vectors scanned per pass");
+DEFINE_int32(reps, 50, "timed repetitions per dim");
+DEFINE_string(dims, "1,2,3,7,16,31,64,128,257",
+              "comma-separated vector dims to benchmark");
+DEFINE_double(tolerance, 1e-4,
+              "max relative diff allowed between SIMD and scalar");
+
+namespace {
+
+// Plain scalar reference -- the ground truth the SIMD kernel must match.
+float ScalarSquaredEuclidean(const float *a, const float *b, size_t dim) {
+  float sum = 0.0f;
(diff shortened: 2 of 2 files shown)


cluster2600:
Closing — opened in error; this wasn't requested. Will revisit #535 when there's an explicit go-ahead.

PR #541, opened 2026-06-30
fix(collection): open LOCK file read-only when collection is read-only

### LINKED ISSUE #6: Read-only collection requires a writable filesystem — `Collection::open` with `read_only=true` fails with `EROFS` on a read-only mount
**Version:** zvec-rust `v0.5.0`

**Summary**

Opening a collection with `read_only` set still opens the collection's `LOCK` file with `O_RDWR`. On a read-only filesystem this fails with `EROFS`, so a read-only collection cannot be opened at all. A genuinely read-only open should never require write access to the storage location.

**Reproduction**

```rust
let mut options = CollectionOptions::new()?;
options.set_read_only(true)?;
options.set_enable_mmap(true)?;
Collection::open(&path, Some(&options))?; // fails if `path` is on a read-only mount
```

```bash
# put a valid collection on a read-only filesystem and open it read-only
mount -o remount,ro /mnt/data
# -> InternalError: Can't open lock file: /mnt/data/<collection>/LOCK
```

In our case this surfaced under a hardened systemd unit (`ProtectSystem=strict`), where the data volume was not in `ReadWritePaths` and was therefore remounted read-only. Every read-only consumer worked except zvec, because zvec is the only component that opens a file `O_RDWR` at open time.

**Root cause**

`CollectionImpl::recovery()` calls `acquire_file_lock(false)`, which opens `LOCK` unconditionally for read-write before taking the (correctly shared) lock:

`vendor/zvec/src/db/collection.cc`, `acquire_file_lock`:

https://github.com/alibaba/zvec/blob/0923f7c6910d46907420e08f6225c82395dab9d3/src/db/collection.cc#L2059-L2073

```cpp
// `create == false` path, taken by recovery()
if (!lock_file_.open(lock_file_path.c_str(), false)) {   // <-- rdonly = false
    return Status::InternalError("Can't open lock file: ", lock_file_path);
}

if (options_.read_only_) {
    if (!ailego::FileLock::TryLockShared(lock_file_.native_handle())) { ... }
} else {
    if (!ailego::FileLock::TryLock(lock_file_.native_handle())) { ... }
}
```
The `false` is the `rdonly` flag of `File::open`, which maps directly to the open flags in `vendor/zvec/src/ailego/io/file.cc`:

```cpp
int flags = rdonly ? O_RDONLY : O_RDWR;
```

So even though the code knows it's a read-only collection (it takes a *shared* lock), the underlying fd is opened `O_RDWR`, which is what fails on a read-only filesystem.

**Suggested fix**

When `options_.read_only_` is set, open the lock file read-only and pair it with the shared lock:

```cpp
bool rdonly = options_.read_only_;
if (!lock_file_.open(lock_file_path.c_str(), rdonly)) {
    return Status::InternalError("Can't open lock file: ", lock_file_path);
}
```

(`flock(LOCK_SH)` works fine on an `O_RDONLY` fd, so the existing `TryLockShared` call is unaffected.)

**Impact**

Read-only collections on read-only media or sandboxed/hardened deployments (read-only bind mounts, immutable images, `systemd` `ProtectSystem=strict`) cannot be opened. Workaround today is to make the storage filesystem writable, which defeats the purpose of `read_only`.

chinaux:
When opening a collection with read_only=true, the LOCK file was unconditionally opened with O_RDWR (POSIX) or GENERIC_WRITE (Windows), causing EROFS/EACCES failures on read-only filesystems.

The fix passes options_.read_only_ to lock_file_.open() so that read-only collections open the LOCK file with:
- POSIX: O_RDONLY instead of O_RDWR
- Windows: GENERIC_READ instead of GENERIC_READ | GENERIC_WRITE

Since flock(LOCK_SH) / shared lock works correctly on read-only file descriptors, the existing shared lock logic remains unaffected.

This enables read-only collections to work on:
- Read-only mounted filesystems (mount -o ro)
- Immutable container images
- Hardened systemd units (ProtectSystem=strict)
- Windows read-only volumes

Added cross-platform unit test Feature_OpenReadOnly_WithReadOnlyLockFile that simulates read-only LOCK file using chmod (POSIX) or SetFileAttributesW (Windows).

Fixes: zvec-ai/zvec-rust#6

### DIFF
--- a/tests/db/collection_test.cc
+++ b/tests/db/collection_test.cc
@@ -23,9 +23,7 @@
 #include <utility>
 #include <vector>
 #include <gtest/gtest.h>
-#if defined(_WIN64) || defined(_WIN32)
-#include <Windows.h>
-#else
+#if !defined(_WIN64) && !defined(_WIN32)
 #include <sys/stat.h>
 #endif
 #include <magic_enum/magic_enum.hpp>
@@ -212,6 +210,14 @@ TEST_F(CollectionTest, Feature_OpenReadOnly_WithReadOnlyLockFile) {
 
 #if defined(_WIN64) || defined(_WIN32)
   // Windows: use SetFileAttributesW to make file read-only
+  // Include Windows.h locally to avoid macro pollution with zvec headers
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#include <Windows.h>
   std::wstring wlock_path(lock_path.begin(), lock_path.end());
   DWORD orig_attrs = GetFileAttributesW(wlock_path.c_str());
   ASSERT_NE(orig_attrs, INVALID_FILE_ATTRIBUTES);
--- a/tests/db/collection_test.cc
+++ b/tests/db/collection_test.cc
@@ -194,6 +194,9 @@ TEST_F(CollectionTest, Feature_CreateAndOpen_General) {
 // This simulates a read-only filesystem scenario (e.g., mount -o ro)
 // See: https://github.com/zvec-ai/zvec-rust/issues/6
 TEST_F(CollectionTest, Feature_OpenReadOnly_WithReadOnlyLockFile) {
+#if defined(_WIN64) || defined(_WIN32)
+  GTEST_SKIP() << "Skipped on Windows: requires chmod";
+#else
   // Create a collection first
   auto schema = TestHelper::CreateNormalSchema();
   auto options = CollectionOptions{false, true, 100 * 1024 * 1024};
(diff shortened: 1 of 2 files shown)


tests/db/collection_test.cc
```diff
@@ -187,6 +190,54 @@ TEST_F(CollectionTest, Feature_CreateAndOpen_General) {
   func(false);
 }
 
+// Test that read-only collection can be opened when LOCK file is read-only
+// This simulates a read-only filesystem scenario (e.g., mount -o ro)
+// See: https://github.com/zvec-ai/zvec-rust/issues/6
+TEST_F(CollectionTest, Feature_OpenReadOnly_WithReadOnlyLockFile) {
+#if defined(_WIN64) || defined(_WIN32)
+  GTEST_SKIP() << "Skipped on Windows: requires chmod";
```
egolearner: use filesystem to support windows

```
#include <filesystem>
#include <system_error>

namespace fs = std::filesystem;

std::error_code ec;
fs::permissions(
    lock_path,
    fs::perms::owner_read | fs::perms::group_read | fs::perms::others_read,
    fs::perm_options::replace,
    ec
);

ASSERT_FALSE(ec) << ec.message();
```

egolearner (changes requested):

tests/db/collection_test.cc
chinaux: ok, fixed.

egolearner (approved):
LGTM

PR #542, opened 2026-06-30
fix(arrow): propagate compiler settings to ExternalProject

### LINKED ISSUE #346: [Bug]: Linker error on Linux x86: multiple definition of std::placeholders::_1
### Description

I am encountering a linker error when building the project on Linux x86 using GCC 12.4. The build fails while linking the shared library libzvec_c_api.so due to a multiple definition error for the symbol std::placeholders::_1.

It appears there is a conflict between the standard library (libstdc++.a) and a symbol already defined in the libarrow.a static library.

### Steps to Reproduce

```python
1. Configure the project on a Linux x86 environment with GCC 12.4.
2. Run the build command and build target all


GCC 13 works fine.
```

### Logs / Stack Trace

```shell
[ 95%] Linking CXX shared library ../../../lib/libzvec_c_api.so
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/12/libstdc++.a(placeholders.o):(.rodata._ZNSt12placeholders2_1E+0x0): multiple definition of `std::placeholders::_1'; ../../../external/usr/local/lib/libarrow.a(type.cc.o):/usr/include/c++/13/functional:294: first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [src/binding/c/CMakeFiles/zvec_c_api.dir/build.make:117: lib/libzvec_c_api.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:3691: src/binding/c/CMakeFiles/zvec_c_api.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

### Operating System

Linux x86

GCC 12.4

### Additional Context

- [x] I built with `CMAKE_BUILD_TYPE=Debug`
- [x] This occurs with or without `COVERAGE=ON`

chinaux:
When the system has multiple GCC versions installed (e.g., GCC 12 and GCC 13), Arrow's ExternalProject could pick up a different compiler or headers than the main project, causing linker errors like:

  multiple definition of `std::placeholders::_1`

The fix explicitly passes CMAKE_C_COMPILER and CMAKE_CXX_COMPILER to Arrow's ExternalProject_Add CONFIGURE_COMMAND on all platforms:
- Linux/macOS
- Windows (MSVC)
- Android
- iOS

This ensures Arrow is built with the same compiler and standard library headers as the main project, preventing symbol conflicts.

Fixes: https://github.com/alibaba/zvec/issues/346


egolearner (approved):
LGTM

egolearner (approved):
LGTM

PR #544, opened 2026-07-02
fix: IVF nprobe selected list scanning

egolearner:
When IVF clusters are skewed, converting an explicit nprobe into an average scan cap can stop scanning after a large early inverted list and skip later selected lists. This makes nprobe behave like a sampled scan budget instead of visiting the closest lists, reducing recall for uneven clustering.

Use the total vector count as the scan cap for explicit nprobe so all selected lists can be fully scanned while keeping scan_ratio behavior unchanged when nprobe is not set.


iaojnh (approved):
LGTM

PR #560, opened 2026-07-08
fix(pyproject.toml): add python_version marker to cibuildwheel and missing pytest-xdist in [dev]

JalinWang:
uv/poetry/pdm resolves dependencies for all Python versions in requires-python (>=3.9), not just the active interpreter. cibuildwheel 3.4.0 requires Python>=3.11, so it must be gated with an environment marker to avoid resolution failures on Python 3.9/3.10."

An example error message:
image


Cuiyus (approved):
LGTM

PR #566, opened 2026-07-08
fix(fts): restore writing segment stats on reopen

### LINKED ISSUE #570: FTS queries can return zero results after reopen because mutable segment stats are not restored
## Problem

FTS queries can return zero results after closing and reopening a collection, even though the documents and postings are still present.

A typical reproduction is:

1. Create a collection with an FTS field.
2. Insert documents.
3. Query in the same process: results are returned.
4. Flush/close the collection without running `optimize()`.
5. Reopen the collection.
6. Run the same FTS query again.

Observed result:

- The query returns zero results after reopen.
- Running `optimize()` before reopen can hide the issue because it rebuilds/seals FTS data into immutable form.

## Root Cause

The writing segment is reopened as a mutable FTS segment. In this path, the FTS postings and document length data can exist on disk, but the in-memory BM25 segment statistics are not restored.

In the old logic, persisted BM25 stats were only loaded for the immutable reader path:

```cpp
// doc_len_cf == nullptr -> immutable path
// doc_len_cf != nullptr -> mutable path
```

For mutable segments, `BM25Scorer` remained at:

```text
total_docs = 0
total_tokens = 0
```

With `total_docs == 0`, BM25 scoring returns zero scores, so FTS queries can effectively return no useful results after reopen.

## Expected Behavior

On reopen:

- Immutable FTS segments must load persisted stats; missing stats should be treated as an error.
- Mutable FTS segments should also try to load persisted stats first.
- If persisted stats exist, restore:
  - scorer stats
  - `total_docs_`
  - `total_tokens_`
- If mutable stats are missing, treat the segment as a brand-new empty mutable segment and keep stats at zero.

## Proposed Fix

Update `FtsColumnIndexer::open_reader()` so it always attempts to load persisted BM25 stats via `load_segment_stats()`.

Then branch by `doc_len_cf` and return value:

```cpp
if load stats failed:
  if doc_len_cf == nullptr:
    // immutable: stats missing is an error
    return error
  else:
    // mutable: new empty segment is allowed
    keep stats at 0
else:
  // restore scorer_, total_docs_, total_tokens_
```

Also ensure `flush()` returns an error if writing segment stats fails, instead of ignoring RocksDB `Put()` failures.

## Verification

Regression coverage should include:

- insert FTS docs
- query before close
- flush/close without optimize
- reopen
- query again
- verify results are still returned

Also cover read-write reopen followed by additional inserts, to ensure restored stats continue from the persisted values.

This sub-issue follows up on #565 and tracks the direct FTS mutable-segment stats restoration bug.

egolearner:
fix #570 

Fix implemented:

- Reopen now always attempts to load persisted segment stats first.
- If `doc_len_cf == nullptr` (immutable path), missing stats is treated as an error.
- If `doc_len_cf != nullptr` (mutable path), missing stats is allowed and treated as a brand-new empty segment.
- `flush()` now checks errors when writing segment stats.
- Added regression coverage for:
  - flush + close + reopen without optimize
  - read-write reopen followed by additional inserts

### DIFF
--- a/python/tests/test_collection_fts.py
+++ b/python/tests/test_collection_fts.py
@@ -20,8 +20,6 @@ optimize all work on a vector-less collection.
 
 from __future__ import annotations
 
-import gc
-
 import pytest
 import zvec
 from zvec import (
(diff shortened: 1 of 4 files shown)


feihongxu0824 (dismissed):

python/tests/test_collection_fts.py
```diff
@@ -168,6 +170,107 @@ def test_optimize_rebuilds_fts(self, fts_collection: Collection):
...
+
+        coll = zvec.create_and_open(
+            path=str(collection_path),
+            schema=schema,
+            option=CollectionOption(read_only=False, enable_mmap=True),
+        )
+        assert all(r.ok() for r in coll.insert(_make_docs()))
+        assert {doc.id for doc in _fts_query(coll, "hello")} == {
+            "pk_0",
+            "pk_1",
```
feihongxu0824: 这个测试文件和c++侧的fts_query_test是什么关系？后续最好都放到c++侧

egolearner: ut已放到c++侧

PR #567, opened 2026-07-08
fix(python): add collection close method

### LINKED ISSUE #563: [Bug]: Python Collection 缺少 close() 方法，导致 Windows 文件锁无法释放
## 问题描述

Zvec Python 绑定的 \Collection\ 对象没有提供 \close()\ 方法。当 collection 被打开后，C++ 底层会持有文件锁（特别是 \scalar.0.ipc\ 等文件），导致：

1. Python 进程退出前无法删除数据目录
2. 无法在同一进程中重新打开同一个目录
3. 在 Windows 上问题尤其严重（文件锁是排他性的）

## 复现步骤

\\\python
import zvec, os, shutil, tempfile

tmp = tempfile.mkdtemp()
    name='test',
    fields=[zvec.FieldSchema(name='x', data_type=zvec.DataType.INT64)],
    vectors=[zvec.VectorSchema(name='v', data_type=zvec.DataType.VECTOR_FP32, dimension=2,
             index_param=zvec.HnswIndexParam(metric_type=zvec.MetricType.COSINE, m=16, ef_construction=200))]
)

col = zvec.create_and_open(tmp, schema)
col.insert(zvec.Doc(id='1', vectors={'v': [1.0, 0.0]}, fields={'x': 1}))
col.optimize()

# 尝试删除数据目录 — 会失败
# 因为没有 close() 方法，C++ 层还持有文件锁
try:
    shutil.rmtree(tmp)
    print('删除成功')
except PermissionError as e:
    print(f'删除失败: {e}')

# 尝试重新打开同一目录 — 也会失败
try:
    col2 = zvec.open(tmp)
    print('重新打开成功')
except Exception as e:
    print(f'重新打开失败: {e}')
\\\

## 预期行为

应有 \close()\ 方法，调用后释放所有底层资源：

\\\python
col.close()
# 之后可以安全删除或重新打开
\\\

## 影响范围

- **Windows 用户**：无法在程序退出前清理数据，文件锁问题导致 shutil.rmtree 失败
- **长期运行的服务**：无法关闭旧 collection 后打开新的
- **测试代码**：每个测试用例后无法清理临时目录

## 临时解决方法

目前只能在 Python 进程退出后手动删除目录，或在进程退出前依赖 GC 回收（但 C++ 层文件锁可能仍未被释放）。

## 建议

在 Python 绑定的 \Collection\ 类中添加 \close()\ 方法，委托给 C++ 层的析构/关闭逻辑。

## 环境

- Zvec v0.5.1（pip 安装）
- Windows 10 x64
- Python 3.11

sjh9714:
## Summary

Closes #563.

Adds `Collection.close()` to release the Python wrapper's underlying `_Collection` handle, so callers can explicitly drop native resources before reopening or removing a collection path. The concurrent-open test now records the opened path before closing the worker handle.

Note: I used Codex to help prepare this change, reviewed the final diff, and ran the listed checks locally.

## Testing

- `git submodule update --init --recursive --depth=1`
- `python3 -m venv .venv`
- `.venv/bin/python -m pip install --upgrade pip setuptools wheel`
- `.venv/bin/python -m pip install -e ".[test]"`
- `.venv/bin/python -m pytest python/tests/detail/test_collection_open.py::TestCollectionOpen::test_close_releases_collection_for_reopen -q`
- `.venv/bin/python -m pytest python/tests/detail/test_collection_open.py -q`
- `.venv/bin/python -m compileall -q python/zvec/model/collection.py python/tests/detail/test_collection_open.py`
- `.venv/bin/python -m pip install "ruff>=0.4"`
- `.venv/bin/python -m ruff check python/zvec/model/collection.py python/tests/detail/test_collection_open.py`
- `git diff --check`


Cuiyus:
Thanks for putting together this PR.
Could we expose an explicit native `Collection::Close()` instead of implementing `close()` by only dropping the Python reference?
`self._obj = None` merely decrements the refcount of the pybind holder. It triggers CollectionImpl destruction only when there are no other references, so an in-flight operation, a shallow copy, or a retained `coll._obj` reference may keep the native collection and its file lock alive after `close()` returns. This is effectively the same behavior already available through del coll when coll is the last reference.

sjh9714:
Thanks for the review — agreed. Dropping the Python reference only decrements a refcount, so it gives no guarantee when other references, shallow copies, or in-flight operations keep the object alive.

I've reworked the PR to expose an explicit native `Collection::Close()`:

- `Collection::Close()` is now a public virtual on the interface, and `CollectionImpl::Close()` reuses the existing `close_unsafe()` teardown (flush for read-write collections, release segments/managers, close the lock file). It takes the exclusive `schema_handle_mtx_`, so it serializes with in-flight operations, and closing twice is a no-op.
- A `closed_` flag is checked alongside the existing `destroyed_` guards, so operations on a closed collection return `InvalidArgument("collection is already closed.")` instead of touching released state through a stale handle. `Destroy()` after `Close()` is also rejected — the intended flow for a closed path is to reopen it (or remove the directory directly, which is what #563 needs on Windows).
- The pybind layer exposes `Close()` (GIL released, same shape as the existing `Destroy` binding), and Python `Collection.close()` now calls it before dropping its references.
- Regression tests: `close()` releases the file lock even while another reference to the native handle is still alive (reopen succeeds), post-close operations through a stale handle raise `ValueError` instead of crashing, and double close is a no-op.

While rerunning the suite I also noticed `test_collection_create_and_open.py::test_open_concurrent_same_path` kept the closed wrapper object and read `.path` on it afterwards; I gave it the same treatment as its twin in `test_collection_open.py` (record the path string before closing).

One note on scope: the C API's `zvec_collection_close` still just deletes its wrapper reference. Happy to wire it to the native `Close()` as a follow-up if you'd like.

**Testing**: `pip install -e ".[dev]"` (full native rebuild) · `pytest python/tests/` — 1107 passed, 70 skipped (platform-only skips) · `ruff check` / `ruff format --check` on the changed Python files

src/db/collection.cc
```diff
@@ -391,6 +400,7 @@ Status CollectionImpl::flush_unsafe() {
 
 Result<std::string> CollectionImpl::Path() const {
   CHECK_DESTROY_RETURN_STATUS_EXPECTED(destroyed_, false);
+  CHECK_CLOSED_RETURN_STATUS_EXPECTED(closed_, false);
```
Cuiyus, line 422: closed_ is a non-atomic bool written by Close() while holding the exclusive schema_handle_mtx_, but Path() reads it without acquiring the same mutex. Concurrent calls to Path() and Close() therefore constitute a C++ data race and undefined behavior.

sjh9714:
Good catch. `Path()` was the only accessor that didn't take `schema_handle_mtx_`, so `closed_` was written under the exclusive lock and read outside it.

Made `closed_` a `std::atomic<bool>` (7e4dd5f). `path_` is set in the constructor and never mutated, so that read stays lock-free.

While here I also fixed the lint failure: the `CHECK_CLOSED_*` macros I added weren't clang-format clean, and that was what the Code Quality job was failing on.

Verified locally: clang-format 18.1.8 and ruff 0.14.4 both clean, and `src/db/collection.cc` compiles with the project's flags. I couldn't run the full test suite — my build directory points at a cmake that no longer exists — so I'm relying on CI for that.

Cuiyus:
@sjh9714  Thanks for the update! This PR now conflicts with the latest `main`, mainly because #614 changed the locking model around `Optimize()`.

Could you please merge the latest `main` into this branch and resolve the conflicts.

Cuiyus (approved):

PR #571, opened 2026-07-08
chore: add clang-tidy pre-commit hook

### LINKED ISSUE #294: clang-tidy: add pre-commit check

Mordris:
Closes #294.

### What changed

- Add a local `clang-tidy` hook to `.pre-commit-config.yaml` so the CI clang-tidy check can be run locally before pushing.
- Add `scripts/run_clang_tidy.py`, which the hook calls. It mirrors `.github/workflows/clang_tidy.yml`: only the changed C/C++ files that appear in `build/compile_commands.json` are analysed, using the repo `.clang-tidy` config and `--warnings-as-errors=*`.
- Files not in the compilation database — and anything under `thirdparty/`, `build/`, or the generated `gen/` dirs — are skipped, matching the workflow's ignore list.
- When clang-tidy or the compilation database is unavailable the hook skips with a hint instead of blocking the commit, so contributors without a configured build are not blocked (CI stays the enforcing gate). Set `ZVEC_CLANG_TIDY_STRICT=1` to make missing tooling an error; `CLANG_TIDY` / `BUILD_DIR` override the binary and build directory.

### How to use

```bash
cmake -S . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
pre-commit install
```

The hook then runs on the C/C++ files you change; commit as usual.

### Validation

- `pre-commit validate-config` passes; `scripts/run_clang_tidy.py` passes the repo `ruff` lint and format checks.
- Verified against a real local build (`cmake -S . -B build -G Ninja -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DBUILD_TOOLS=ON`; compilation database with ~1.6k entries; clang-tidy 18.1.1), driven through `pre-commit run clang-tidy`:
  - clean upstream sources (e.g. files under `src/ailego/`) pass;
  - a `modernize-use-nullptr` violation temporarily added to a tracked source file makes the hook fail (exit 1) with the diagnostic, and passes again once reverted;
  - files under `thirdparty/` and generated `gen/` dirs, and files absent from the compilation database, are skipped;
  - with no `build/compile_commands.json` the hook skips without blocking the commit (and fails under `ZVEC_CLANG_TIDY_STRICT=1`).


egolearner:
Thanks a lot for this contribution, @Mordris — solid foundation for wiring clang-tidy into the local workflow! 🙏

I've pushed a follow-up commit on top of your work to align it with the direction we'd like to take here

1. Runs at the pre-push stage instead of on every commit, so day-to-day commits stay fast while pushes still get checked.
2. Pins clang-tidy==18.1.8 via pre-commit's managed virtualenv (language: python + additional_dependencies) to match the version used in the CI clang_tidy workflow.
3. Flexible compile_commands.json discovery: repo root wins outright, otherwise $BUILD_DIR / the well-known build, build.release, build.debug dirs are tried, falling back to a shallow scan of the working tree (up to two levels).

.pre-commit-config.yaml
```diff
@@ -44,6 +45,23 @@ repos:
...
+        name: clang-tidy (C++)
+        description: >-
+          Static analysis on changed C/C++ files, mirroring the CI clang-tidy
+          check. Uses the repo .clang-tidy config and the CMake compilation
+          database (build/, build.release/, build.debug/, or $BUILD_DIR); files
+          not in the database are skipped. Configure a build with
+          `cmake -S . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON` to enable it.
+        entry: scripts/run_clang_tidy.py
+        language: python
+        additional_dependencies: [clang-tidy==18.1.8]
```
richyreachy, line 58: The clang-tidy==18.1.8 package doesn’t provide a prebuilt wheel for Linux ARM64, so pre-commit may try to build LLVM from source before the script’s graceful-skip logic can run. A system-installed clang-tidy-18 or clang-tidy could be used. As well, it would be nice to skip with a helpful message if neither is available.

richyreachy (approved):

.pre-commit-config.yaml
egolearner: Thanks for the information. We'll update clang-tidy version in another PR.

PR #584, opened 2026-07-10
fix(fts): preserve zero-match filter semantics

### LINKED ISSUE #583: [Bug]: FTS/hybrid query drops an indexed scalar filter when the filter matches zero documents
### Description

When a query combines a full-text (FTS) match with a scalar filter, and that filter is backed by an inverted index (`InvertIndexParam`) and matches **zero** documents, the filter is silently dropped and the query returns every FTS hit. The same is true for hybrid queries (FTS + vector). The scalar-only query path (`query(filter=..., topk=...)` with no FTS/vector part) is correct and returns an empty result for the same filter.

The trigger is the filter selecting zero rows, not anything specific to strings or to a value being "unknown". It reproduces with string equality, integer equality, and integer range predicates, as long as the field has an inverted index. If the filter matches at least one document, it is applied correctly, even when the matched set is disjoint from the FTS hits (the result is then correctly empty). If the field has no inverted index, the filter is also applied correctly, because that path uses a different (forward) evaluator.

Expected: FTS query + `status = 'archived'` returns `[]` when no document has status `archived`.
Actual: it returns every document that matches the FTS query.

I originally hit this through a filtered FTS search that started returning everything after I deleted the last document with a given status. Deletion was incidental: it just made the filter match zero rows, which is the real trigger.

### Steps to Reproduce

```python
import tempfile
import zvec
from zvec import (
    CollectionSchema, FieldSchema, VectorSchema, CollectionOption, Doc,
    DataType, MetricType, HnswIndexParam, InvertIndexParam, FtsIndexParam,
    Query, Fts, OptimizeOption,
)

path = tempfile.mkdtemp() + "/coll"
schema = CollectionSchema(
    name="ftsfilter",
    fields=[
        FieldSchema("status", DataType.STRING, nullable=False, index_param=InvertIndexParam()),
        FieldSchema("views", DataType.INT64, nullable=False, index_param=InvertIndexParam()),
        FieldSchema("body", DataType.STRING, nullable=False,
                    index_param=FtsIndexParam(tokenizer_name="standard", filters=["lowercase"])),
    ],
    vectors=[VectorSchema("dense", DataType.VECTOR_FP32, dimension=4,
                          index_param=HnswIndexParam(metric_type=MetricType.COSINE))],
)
c = zvec.create_and_open(path=path, schema=schema, option=CollectionOption(read_only=False))
c.insert([  # every body contains "hello"; status draft/published/published; views 10/20/30
    Doc(id="1", fields={"status": "draft",     "views": 10, "body": "hello one"},   vectors={"dense": [1.0, 0, 0, 0]}),
    Doc(id="2", fields={"status": "published", "views": 20, "body": "hello two"},   vectors={"dense": [0, 1.0, 0, 0]}),
    Doc(id="3", fields={"status": "published", "views": 30, "body": "hello three"}, vectors={"dense": [0, 0, 1.0, 0]}),
])
c.optimize(option=OptimizeOption())

def fts(filt):
    q = Query(field_name="body", fts=Fts(match_string="hello"))
    return sorted(x.id for x in c.query(queries=q, topk=10, filter=filt))

def scalar(filt):
    return sorted(x.id for x in c.query(topk=10, filter=filt))

for filt in ["status = 'archived'", "views = 999", "views > 1000", "status = 'published'"]:
    print(f"{filt:22s} FTS={str(fts(filt)):16s} scalar-only={scalar(filt)}")
```

Output:

```
status = 'archived'    FTS=['1', '2', '3']  scalar-only=[]
views = 999            FTS=['1', '2', '3']  scalar-only=[]
views > 1000           FTS=['1', '2', '3']  scalar-only=[]
status = 'published'   FTS=['2', '3']       scalar-only=['2', '3']
```

The first three filters match zero documents; the FTS path returns all three docs while the scalar-only path correctly returns `[]`. `optimize()` and `flush()` do not change the outcome.

### Further observations

Behavior across operators and types (field has an inverted index; FTS term "hello" matches all three docs):

```
filter                 matches   FTS path         scalar-only
status = 'draft'          1      ['1']            ['1']
status = 'archived'       0      ['1','2','3']    []            <- filter dropped
views  = 20               1      ['2']            ['2']
views  = 999              0      ['1','2','3']    []            <- filter dropped
views  > 25               1      ['3']            ['3']
views  > 1000             0      ['1','2','3']    []            <- filter dropped
status != 'archived'      3      ['1','2','3']    ['1','2','3']
views  > 5                3      ['1','2','3']    ['1','2','3']
```

- Hybrid queries (FTS + vector with `RrfReRanker`) show the same drop for the zero-match filters.
- Independent of segment state: reproduces with no `optimize()`, with `optimize()`, and with multiple sealed segments.
- Reproduces with the filtered field both `nullable=False` and `nullable=True`, and with no null values present (so this is unrelated to #409).
- If the filter matches at least one document but is disjoint from the FTS hits, the result is correctly empty. For example, FTS matching only doc 1 (`draft`) filtered by `status = 'published'` (docs 2 and 3) returns `[]`.
- Discriminator: the same zero-match filter on a field that has **no** inverted index returns `[]` correctly:

```
WITH    inverted index, FTS + status='archived' (0 match): ['1', '2', '3']
WITHOUT inverted index, FTS + status='archived' (0 match): []
```

### Root cause

Traced by reading the source (paths as of v0.5.1; behavior reproduced on the 0.5.1 PyPI wheel). A zero-match inverted search is represented by a null bitmap, and downstream code treats "null/empty inverted result" as "no filter":

- `InvertedSearchResult::make_filter()` in `src/db/index/column/inverted_column/inverted_search_result.h` returns `nullptr` when the internal roaring bitmap is null. A zero-match search leaves the bitmap null (the in-place `AND` frees the bitmap whenever either operand is empty).
- `DocFilter::empty()` in `src/db/sqlengine/planner/doc_filter.cc` returns true when none of `delete_filter_ / invert_filter_ / forward_plan_ / forward_filter_expr_` is set. For an indexed field there is no forward expression, so a null `invert_filter_` makes the whole filter look empty.
- `FtsRecallNode::prepare()` in `src/db/sqlengine/planner/fts_recall_node.cc` then does `params.filter = doc_filter_->empty() ? nullptr : doc_filter_;`, i.e. runs FTS recall with no filter. In the same function, `get_bf_by_keys_and_update()` returns an empty candidate-id list for the zero-match result, and `FtsColumnIndexer` (`src/db/index/column/fts_column/fts_column_indexer.cc`) only applies the candidate restriction when the list is non-empty, so that restriction is dropped too.
- `VectorRecallNode` (`src/db/sqlengine/planner/vector_recall_node.cc`) has the same `doc_filter_->empty() ? nullptr : ...` line, which is why hybrid queries are affected.
- The scalar-only path (`InvertRecallNode`, `src/db/sqlengine/planner/invert_recall_node.cc`) uses the inverted result as its candidate source rather than as a filter, so a null bitmap yields an invalid iterator and the correct empty result. That is why only the FTS/hybrid paths misbehave.

The behavioral discriminator above (field with vs without an inverted index) matches this: the no-index case uses the forward evaluator, so `DocFilter::empty()` is false and the filter is not dropped.

One possible fix direction is to distinguish "no filter" from "filter that matched zero documents", for example by representing a zero-match inverted result with an empty non-null bitmap so `make_filter()` returns a real reject-all filter, or by having the recall nodes treat a computed-but-empty inverted result as reject-all instead of as absent.

### Logs / Stack Trace

No crash or error; the query returns the wrong rows (see the outputs above).

### Operating System

Ubuntu (Linux 5.15.0, x86_64)

Reproduced two ways, same wrong output both times:
- zvec 0.5.1 (PyPI wheel), Python 3.14.2.
- From source at `v0.5.1` (working tree `v0.5.1-22`, with the `sqlengine` planner identical to `v0.5.1`), Python 3.14, built with GCC 11 (GCC 15 cannot compile bundled RocksDB 8.1.1), Ubuntu 22.04 x86_64. Built both `CMAKE_BUILD_TYPE=Debug` and the `COVERAGE` config (`--coverage`); the FTS-path filter drop is present in both.

### Additional Context

- [x] I built with `CMAKE_BUILD_TYPE=Debug`
- [x] This occurs with or without `COVERAGE=ON`
- [ ] The issue involves Python ↔ C++ integration (pybind11) — the defect is in the C++ query planner (`sqlengine`); the Python binding only passes the filter string through

egolearner:
fix #583

### DIFF
--- a/src/db/index/column/fts_column/fts_column_indexer.cc
+++ b/src/db/index/column/fts_column/fts_column_indexer.cc
@@ -206,11 +206,14 @@ Result<std::vector<FtsResult>> FtsColumnIndexer::search(
   // Candidate-driven mode: AND a CandidateDocIterator into the root so the
   // small candidate set leads (Conjunction sorts by cost asc), turning the
   // posting walk into per-candidate advance()+matches()+score().
-  if (!query_params.candidate_ids.empty()) {
+  if (query_params.candidate_ids) {
+    if (query_params.candidate_ids->empty()) {
+      return std::vector<FtsResult>{};
+    }
     std::vector<DocIteratorPtr> musts;
     musts.reserve(2);
     musts.push_back(
-        std::make_unique<CandidateDocIterator>(query_params.candidate_ids));
+        std::make_unique<CandidateDocIterator>(*query_params.candidate_ids));
     musts.push_back(std::move(root_iter));
     root_iter = std::make_unique<ConjunctionIterator>(
         std::move(musts), std::vector<DocIteratorPtr>{});
--- a/src/db/index/column/fts_column/fts_types.h
+++ b/src/db/index/column/fts_column/fts_types.h
@@ -15,6 +15,7 @@
 #pragma once
 
 #include <cstdint>
+#include <optional>
 #include <string>
 #include <vector>
 #include "db/index/common/index_filter.h"
@@ -28,11 +29,11 @@ struct FtsQueryParams {
   // Wraps zvec::IndexFilter for push-down filtering inside the search loop.
   IndexFilter::Ptr filter{nullptr};
   // Candidate-driven (brute-force) mode: ascending segment-local doc_ids;
-  // when non-empty, FtsColumnIndexer restricts evaluation to this set by
-  // AND-ing it with the root iterator. Filled by the planner via
+  // nullopt means no candidate restriction, while a present empty vector
+  // means no document can match. Filled by the planner via
   // DocFilter::get_bf_by_keys_and_update when an invert result is highly
   // selective.
-  std::vector<uint64_t> candidate_ids;
+  std::optional<std::vector<uint64_t>> candidate_ids;
 };
 
 /*! Per-segment statistics needed by the FTS reducer for doc_id remapping.
--- a/tests/db/index/column/fts_column/fts_column_indexer_test.cc
+++ b/tests/db/index/column/fts_column/fts_column_indexer_test.cc
@@ -1778,7 +1778,8 @@ TEST_F(FtsColumnIndexerTest, BruteForceCoexistsWithFilterPushdown) {
 
   zvec::fts::FtsQueryParams qp;
   qp.topk = 10;
-  qp.candidate_ids = {0, 1, 2};          // candidates restrict to {0,1,2}
+  qp.candidate_ids =
+      std::vector<uint64_t>{0, 1, 2};    // candidates restrict to {0,1,2}
   qp.filter = make_blocked_filter({1});  // further drop doc 1
   auto ret = indexer->search(*ast, qp);
   ASSERT_TRUE(ret.has_value());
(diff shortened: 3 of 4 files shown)


JalinWang (approved):

egolearner:
verified with https://github.com/egolearner/zvec/pull/5

PR #600, opened 2026-07-17
fix: optimize after crash recovery without opening wal

### LINKED ISSUE #594: [Bug]: optimize() silently discards WAL-replayed data on a crash-recovered Collection
### Description

We evaluated ZVec's in-process storage as a durable backing store for a workflow orchestrator (documents inserted via `Collection.insert()`, no explicit `flush()` call before an unclean process termination). Raw document durability across a hard crash (`os._exit(1)`, equivalent to `kill -9`) is solid: `Collection.stats().doc_count` is fully and correctly restored on `zvec.open()` via automatic WAL replay, with no manual recovery call required.

However, calling `Collection.optimize()` on a collection that carries this crash-recovered history causes the WAL-replayed documents to be silently dropped from the persisted collection. The call itself completes without raising an exception, and `stats()` on the *same* handle right after `optimize()` reports the pre-existing `doc_count` unchanged together with `index_completeness: 1.0` — appearing to be a successful, lossless index build. Only when the collection is reopened from a **separate process** does the loss become visible: `doc_count` regresses to whatever was last durably segmented to disk before the crash (in our minimal repro below, from 150 to 0, since no `flush()` was ever called).

We isolated the exact trigger through a control experiment: a collection that has **unflushed data but no crash history** (created, populated, and shut down normally, no `os._exit`) retains 100% of its documents after `optimize()`, including the unflushed ones. The data loss is therefore not a general property of "unflushed data" — it is specific to documents that were recovered via WAL replay from a genuine unclean termination. `optimize()`'s index-rebuild path appears to exclude that WAL-replayed data from the rebuilt segment rather than merging it in.

We did not find any signal in the public API that a collection carries this crash-recovered state at the time `optimize()` is called — no field on `stats()`, no exception, no return value from `optimize()` itself. The only observable trace is a one-time `possible crash residue; cleaning and overwriting` warning logged to stderr the first time the collection is reopened after the crash; it does not persist as a queryable flag afterward.

**Expected behavior:**

`optimize()` should complete the index for a crash-recovered collection without discarding the documents that WAL replay already restored on `open()`. At minimum, if some subset of WAL-replayed data cannot be safely folded into the optimized index, `optimize()` should surface that (an exception, a non-1.0 completeness value, or a count in the return value) rather than reporting `index_completeness: 1.0` while having silently dropped documents that `stats()` reported as present moments earlier.

**Measured Workaround:**

Rather than repairing the crash-recovered collection in place, migrating its contents into a freshly created collection (which carries no crash history) before calling `optimize()` avoided the loss entirely in our testing. Verified against a 551-document crash-recovered collection: `fetch()` retrieved 551/551 documents with zero missing, the reinsert into the fresh collection preserved all vectors byte-for-byte (`np.allclose` against the originals), and the subsequent `optimize()` on the fresh collection produced `doc_count: 551` and `index_completeness: {'embedding': 1.0}` on a separate-process reopen, with `query()` returning correctly ranked results.

```python
import zvec

# STEP 1 — open the crashed collection. This alone triggers WAL replay;
# doc_count comes back correct even though index_completeness is still 0.0.
crashed = zvec.open("./zvec_crash_test")
print(crashed.stats)
# doc_count=<N> index_completeness={'embedding': 0.0}

# STEP 2 — fetch every document by known ID. This step depends on already
# knowing the full ID space; zvec has no full-collection-scan / list-all-ids
# primitive that we could find, so an external ID ledger is required for
# collections whose ID space isn't a known, predictable range.
all_ids = [str(i) for i in range(N)]
docs = crashed.fetch(all_ids, include_vector=True)
assert len(docs) == N

# STEP 3 — create a brand-new collection. It has no crash history, so it
# is not subject to the exclusion behavior described above.
recovered = zvec.create_and_open("./zvec_recovered", schema)

# STEP 4 — reinsert through the normal insert() path, not a raw file copy.
recovered.insert(list(docs.values()))
recovered.flush()

# STEP 5 — only now, on the crash-free collection, call optimize().
recovered.optimize()
# Verified on a fresh reopen: doc_count == N, index_completeness == 1.0,
# no documents missing.
```

This is a workaround, not a fix: it depends on the caller already knowing every document ID in the crashed collection, which `zvec`'s public API does not expose a way to enumerate. We are filing this issue primarily for the `optimize()` data-loss behavior described above; the missing enumeration/full-scan primitive may be worth tracking as a separate, related issue if useful.

### Steps to Reproduce

```python
Save the script below as `repro_zvec_optimize_dataloss.py` and run `python3 repro_zvec_optimize_dataloss.py` in an empty working directory. It is self-contained: it creates `./zvec_crash_test` itself (relative path, no fixtures needed) and drives every phase (crash, reopen, optimize, reopen) in its own subprocess, so each `stats()` read is genuinely fresh and not a stale in-process cache.


"""
Minimal reproduction: optimize() silently discards WAL-replayed data on a
crash-recovered zvec Collection.

Run: python3 repro_zvec_optimize_dataloss.py

The script has four phases, each executed in its OWN process via subprocess
(a same-process handle can report stale/cached stats, so every check here
is a genuinely fresh `zvec.open()` in a new OS process):

  [1] Insert NUM_DOCS_BEFORE_CRASH docs with no flush() calls, then
      os._exit(1) mid-run in a child process (simulates kill -9).
  [2] Fresh process re-opens the collection and reports stats() -- this
      is expected to show full recovery via WAL replay.
  [3] Fresh process re-opens the collection and calls optimize().
  [4] Fresh process re-opens the collection again and reports stats() --
      this is where the data loss becomes visible.
"""

import os
import shutil
import subprocess
import sys

import numpy as np
import zvec

COLLECTION_PATH = "./zvec_crash_test"
DIM = 64
NUM_DOCS_BEFORE_CRASH = 150  # 0-indexed ids 0..149 will be inserted, no flush() at all


def build_schema() -> zvec.CollectionSchema:
    id_field = zvec.FieldSchema("id", zvec.DataType.INT64, nullable=False)
    embedding_field = zvec.VectorSchema(
        "embedding",
        data_type=zvec.DataType.VECTOR_FP32,
        dimension=DIM,
        index_param=zvec.HnswIndexParam(metric_type=zvec.MetricType.COSINE),
    )
    return zvec.CollectionSchema(name="repro_collection", fields=id_field, vectors=embedding_field)


def phase_crash_child() -> None:
    """Runs in a dedicated child process. Inserts docs with no flush(), then os._exit(1)."""
    rng = np.random.default_rng(seed=1)
    schema = build_schema()
    collection = zvec.create_and_open(COLLECTION_PATH, schema)
    for i in range(NUM_DOCS_BEFORE_CRASH):
        vec = rng.random(DIM, dtype=np.float32).tolist()
        doc = zvec.Doc(id=str(i), fields={"id": i}, vectors={"embedding": vec})
        collection.insert(doc)
    print(f"  [child] inserted {NUM_DOCS_BEFORE_CRASH} docs, no flush() called, now os._exit(1)", flush=True)
    os._exit(1)  # kill -9 equivalent: no atexit, no destructors, no flush


def phase_check(label: str) -> None:
    """Runs in a dedicated fresh process. Opens the collection and prints stats."""
    collection = zvec.open(COLLECTION_PATH)
    s = collection.stats
    print(f"  [{label}] doc_count={s.doc_count} index_completeness={s.index_completeness}", flush=True)


def phase_optimize() -> None:
    """Runs in a dedicated fresh process. Opens the collection and calls optimize()."""
    collection = zvec.open(COLLECTION_PATH)
    before = collection.stats
    print(f"  [optimize] before: doc_count={before.doc_count} index_completeness={before.index_completeness}", flush=True)
    collection.optimize()
    after = collection.stats
    print(f"  [optimize] same-handle after optimize(): doc_count={after.doc_count} index_completeness={after.index_completeness}", flush=True)


def main() -> None:
    if os.path.exists(COLLECTION_PATH):
        shutil.rmtree(COLLECTION_PATH)

    print("[1] Inducing a hard crash (os._exit(1), zero flush() calls) in a child process...", flush=True)
    subprocess.run([sys.executable, __file__, "--crash-child"])

    print("[2] Fresh process re-open right after the crash (expected: full WAL-replay recovery):", flush=True)
    subprocess.run([sys.executable, __file__, "--check", "after-crash"])

    print("[3] Fresh process re-open, then optimize() on the crash-recovered collection:", flush=True)
    subprocess.run([sys.executable, __file__, "--optimize"])

    print("[4] Fresh process re-open AFTER optimize() (this is where the loss is visible):", flush=True)
    subprocess.run([sys.executable, __file__, "--check", "after-optimize"])


if __name__ == "__main__":
    if "--crash-child" in sys.argv:
        phase_crash_child()
    elif "--optimize" in sys.argv:
        phase_optimize()
    elif "--check" in sys.argv:
        phase_check(sys.argv[-1])
    else:
        main()
```

### Logs / Stack Trace

```shell
Verified output from running the script above, unmodified, in an empty directory:


[1] Inducing a hard crash (os._exit(1), zero flush() calls) in a child process...
  [child] inserted 150 docs, no flush() called, now os._exit(1)
[2] Fresh process re-open right after the crash (expected: full WAL-replay recovery):
[ WARN ... segment.cc:4128] ForwardBlock file[./zvec_crash_test/0/scalar.0.ipc] already exists (possible crash residue); cleaning and overwriting.
[ WARN ... segment.cc:4101] Index file[./zvec_crash_test/0/embedding.index.1.proxima] already exists (possible crash residue); cleaning and overwriting.
  [after-crash] doc_count=150 index_completeness={'embedding': 0.0}
[3] Fresh process re-open, then optimize() on the crash-recovered collection:
[ WARN ... segment.cc:4128] ForwardBlock file[./zvec_crash_test/0/scalar.0.ipc] already exists (possible crash residue); cleaning and overwriting.
[ WARN ... segment.cc:4101] Index file[./zvec_crash_test/0/embedding.index.1.proxima] already exists (possible crash residue); cleaning and overwriting.
  [optimize] before: doc_count=150 index_completeness={'embedding': 0.0}
  [optimize] same-handle after optimize(): doc_count=150 index_completeness={'embedding': 1.0}
[4] Fresh process re-open AFTER optimize() (this is where the loss is visible):
  [after-optimize] doc_count=0 index_completeness={'embedding': 0.0}


Note on `index_completeness: 0.0` in step `[after-crash]` and again in the final `[after-optimize]` line: this value alone is not itself an anomaly — it is also the default, harmless state of any collection (crashed or not) before `optimize()` has ever been called on it, and `query()` still returns correct results while it holds. It is included here only for completeness of the raw output; the actionable defect is the `doc_count` regression from `150` to `0` between steps `[3]` and `[4]`, which happens purely as a result of calling `optimize()`.

We additionally confirmed, in a separate run (not included in this minimal script for brevity), that a collection built with a periodic `flush()` cadence loses only the documents inserted after the last `flush()` rather than the entire dataset — e.g., with 551 documents inserted and `flush()` called every 100 documents, `optimize()` on the crash-recovered collection left 501 documents (IDs 501–550, all inserted after the last `flush()` at ID 500, were missing on `fetch()`). This indicates the loss boundary is exactly the last durably-flushed segment, and that the mechanism is specific to WAL-replayed (crash-recovered) data rather than "unflushed data" in general — a control collection with unflushed data but **no** crash history retained 100% of its documents after `optimize()`.
```

### Operating System

macOS 26.5.1 (Apple Silicon, arm64)

zvec 0.5.1 (PyPI wheel, no source build), Python 3.13.8, numpy 2.5.1

### Additional Context

feihongxu0824:
## Summary

- keep the recovered WAL attached to the segment after replay
- ensure `Optimize()` flushes data restored from the WAL before sealing the segment
- add a crash-recovery regression test covering optimize, close, and reopen

## Root cause

Segment recovery replayed WAL records into memory using a local WAL handle, then closed it without restoring the member WAL handle. `Optimize()` subsequently treated the writing segment as having no WAL records and skipped persistence, so recovered documents disappeared after reopening.

## Impact

Recovered documents are now persisted when a collection is optimized immediately after crash recovery.

Fixes #594.

## Validation

- all 6 `CrashRecoveryTest` cases pass locally
- full CI matrix passed on the same commit: Linux, macOS, Windows, Android, iOS, lint, and clang-tidy
- `git diff --check` passes

### DIFF
--- a/src/db/index/segment/segment.cc
+++ b/src/db/index/segment/segment.cc
@@ -4212,8 +4212,6 @@ Status SegmentImpl::recover() {
               wal_file_path.c_str());
     return Status::OK();
   }
-  AILEGO_DEFER([&]() { recover_wal_file->close(); });
-
   std::array<uint64_t, static_cast<size_t>(Operator::DELETE) + 1>
       recovered_doc_count{};
   uint64_t total_recovered_doc_count{0};
@@ -4297,7 +4295,17 @@ Status SegmentImpl::recover() {
       (size_t)recovered_doc_count[3]   // DELETE
   );
 
-  return Status::OK();
+  if (recover_wal_file->close() != 0) {
+    return Status::InternalError("Failed to close recovered wal file: ",
+                                 wal_file_path);
+  }
+  recover_wal_file.reset();
+
+  // Keep the recovered WAL attached to the segment. Operations such as
+  // optimize() flush the writing segment before sealing it; without an open
+  // member WAL, flush() treats the recovered memory components as empty and
+  // returns without persisting them.
+  return open_wal_file();
 }
 
 Status SegmentImpl::open_wal_file() {
--- a/tests/db/crash_recovery/write_recovery_test.cc
+++ b/tests/db/crash_recovery/write_recovery_test.cc
@@ -218,6 +218,39 @@ TEST_F(CrashRecoveryTest, CrashRecoveryDuringInsertion) {
 }
 
 
+TEST_F(CrashRecoveryTest, OptimizeAfterCrashRecoveryPersistsReplayedData) {
+  {
+    auto schema = CreateTestSchema(collection_name_);
+    auto result = Collection::CreateAndOpen(dir_path_, *schema, options_);
+    ASSERT_TRUE(result.has_value()) << result.error().message();
+  }
+
+  RunGeneratorAndCrash("0", "10000", "insert", "0", 1);
+
+  uint64_t recovered_doc_count = 0;
+  {
+    auto result = Collection::Open(dir_path_, options_);
+    ASSERT_TRUE(result.has_value())
+        << "Failed to reopen collection after crash recovery";
+    auto collection = result.value();
+
+    recovered_doc_count = collection->Stats().value().doc_count;
+    ASSERT_GT(recovered_doc_count, 0)
+        << "No documents were recovered from the WAL";
+
+    auto status = collection->Optimize();
+    ASSERT_TRUE(status.ok()) << status.message();
+    ASSERT_EQ(collection->Stats().value().doc_count, recovered_doc_count);
+  }
+
+  auto result = Collection::Open(dir_path_, options_);
+  ASSERT_TRUE(result.has_value())
+      << "Failed to reopen collection after optimizing recovered data";
+  ASSERT_EQ(result.value()->Stats().value().doc_count, recovered_doc_count)
+      << "Optimize discarded documents restored from the WAL";
+}
+
+
 TEST_F(CrashRecoveryTest, CrashRecoveryDuringUpsert) {
   {
     auto schema = CreateTestSchema(collection_name_);


feihongxu0824:
verified with https://github.com/feihongxu0824/zvec/pull/8

chinaux (approved):
LGTM

PR #612, opened 2026-07-22
fix(python): validate query field names

HosniBelfeki:
## Summary

Tighten Python Query validation so field_name must be a non-empty string.

Previously, Query._validate() only rejected None, so an empty string or non-string value could pass Python-side validation and fail later with a less direct executor/schema error.

This adds a focused regression test for None, empty string, and non-string field names.

## Tests

- python -m ruff check python\zvec\model\param\query.py python\tests\test_params.py
- python -m py_compile python\zvec\model\param\query.py python\tests\test_params.py

### DIFF
--- a/python/tests/test_params.py
+++ b/python/tests/test_params.py
@@ -408,9 +408,10 @@ class TestQuery:
         with pytest.raises(ValueError):
             Query(field_name="embedding", id="doc123", vector=[0.1])._validate()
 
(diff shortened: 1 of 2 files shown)


HosniBelfeki:
Addressed Copilot review feedback: whitespace-only `field_name` values are now rejected with `field_name.strip()`, and the regression test covers whitespace-only input.

Cuiyus (approved):

PR #615, opened 2026-07-23
fix(diskann): make Linux AIO fallback safe for partial operations

iaojnh:
# Summary

Fix DiskANN’s Linux file reader to safely handle partial libaio submissions and completions.

Previously, any result smaller than the requested batch size was treated as a failure and immediately fell back to pread(). When AIO requests were still in flight, both paths could write to the same destination buffers concurrently, potentially causing races or data corruption.

# Changes

 - Accumulate partial io_submit() results and submit only the remaining requests.

 - Drain every submitted AIO request before falling back to synchronous reads.

 - Continue waiting after EINTR while requests are still in flight.

 - Reset the AIO context after a fatal io_getevents() failure before using pread().

 - Match completions to requests through io_event::obj instead of assuming completion order.

 - Validate res, res2, duplicate events, and unknown completions.

 - Fall back to pread() when the AIO context is unavailable, and make context teardown null-safe.

# Test coverage

Added Linux-specific unit tests covering:

 - Partial submissions and partial completions.

 - Submission failure after part of a batch has already been accepted.

 - Draining all outstanding completions before falling back after a short read.


richyreachy (approved):
LGTM

PR #616, opened 2026-07-23
fix(python): validate query topk

HosniBelfeki:
## Summary

Validate `Collection.query(topk=...)` before building the query context.

The Python wrapper already has `_require_positive_integer()` for group-by limits, but plain `query()` passed invalid `topk` values through to the executor/native layer. This change rejects `0`, negative values, `None`, and booleans with a clear Python-side `ValueError`.

## Tests

- `python -m ruff check python\zvec\model\collection.py python\tests\test_collection.py`
- `python -m py_compile python\zvec\model\collection.py python\tests\test_collection.py`

### DIFF
--- a/python/tests/test_collection.py
+++ b/python/tests/test_collection.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 from __future__ import annotations
 
+from unittest.mock import MagicMock
+
 import pytest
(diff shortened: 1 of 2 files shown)


Cuiyus (approved):
LGTM

PR #623, opened 2026-07-28
refactor(thread-pool): make CPU affinity opt-in

luoxiaojian:
Make thread counts explicit and cap worker creation at hardware concurrency. Respect Linux affinity masks, fix cgroup v2 CPU quota parsing, and default DB query and optimize pools to no CPU binding without exposing affinity controls through public C or Python APIs.


feihongxu0824 (approved):
LGTM

PR #640, opened 2026-08-02
fix(python): reject empty fts queries

HosniBelfeki:
Summary

This hardens Python FTS query validation by rejecting Fts objects that do not provide a non-empty query_string or match_string. Previously, empty or whitespace-only FTS values could pass Python-side validation and reach the native query path.

The executor now forwards stripped FTS text to the native _Fts object, keeping validation and execution behavior consistent.

Tests

- python -m ruff check python/zvec/model/param/query.py python/zvec/executor/query_executor.py python/tests/test_fts_query.py python/tests/test_query_executor.py
- python -m py_compile python/zvec/model/param/query.py python/zvec/executor/query_executor.py python/tests/test_fts_query.py python/tests/test_query_executor.py

Note: focused pytest against local source is blocked in this workspace because zvec._zvec is not built locally.


python/zvec/model/param/query.py
```diff
@@ -111,9 +111,19 @@ def has_fts(self) -> bool:
             bool: True if `fts` is set with a query_string or match_string.
         """
         if self.fts is not None:
-            return bool(self.fts.query_string) or bool(self.fts.match_string)
+            return bool(self._fts_query_string()) or bool(self._fts_match_string())
         return False
 
+    def _fts_query_string(self) -> str:
```
Cuiyus, line 129: Calling `strip()` here rewrites every query before it reaches the native tokenizer. Consider this concrete example:

```python
document = "\u00a0foo"  # UTF-8: b"\xc2\xa0foo"
query = "\u00a0foo"
```

During indexing, `WhitespaceTokenizer` checks whitespace byte-by-byte with `std::isspace`, so the UTF-8 NBSP remains part of the indexed token:

```text
indexed token: "\u00a0foo"
```

Before this change, the same query produced the same token and matched. After this change:

```python
query.strip()  # "foo"
```

the query token becomes `"foo"` and no longer matches `"\u00a0foo"`, silently returning no results.

Please use `strip()` only to detect all-whitespace values while preserving the original non-empty text, or apply identical normalization during indexing. An end-to-end regression test using `WhitespaceTokenizer` and this boundary-NBSP example would cover the issue.

HosniBelfeki:
Updated the FTS validation path so strip() is only used to detect empty or whitespace-only input. Valid query text is now passed through unchanged, including boundary whitespace/NBSP characters, and non-string query_string/match_string values raise a clear ValueError during validation. I also updated the executor regression test to assert preservation rather than trimming.

Cuiyus (approved):
LGTM

PR #643, opened 2026-08-02
fix(python): validate collection document ids

HosniBelfeki:
Summary

This adds Python-side normalization and validation for document IDs passed to Collection.fetch() and Collection.delete().

Previously, invalid values such as None, bytes, non-string values, or lists containing empty/non-string IDs could reach the native layer and fail with less direct errors. This now rejects invalid IDs early with a clear ValueError, while still preserving single-ID return behavior and accepting string sequences such as tuples by normalizing them to lists.

Tests

- python -m ruff check python/zvec/model/collection.py python/tests/test_collection.py
- python -m ruff format --check python/zvec/model/collection.py python/tests/test_collection.py
- python -m py_compile python/zvec/model/collection.py python/tests/test_collection.py

Note: focused pytest in this local workspace resolves the installed zvec package instead of the edited source package, so the new tests are intended to run in CI against the PR source tree.

### DIFF
--- a/python/tests/test_collection.py
+++ b/python/tests/test_collection.py
@@ -426,6 +426,26 @@ class TestCollectionOptimize:
 # ----------------------------
 @pytest.mark.usefixtures("test_collection")
 class TestCollectionFetch:
+    @pytest.mark.parametrize("ids", [None, 1, b"doc_1", "", ["doc_1", None], [""]])
+    def test_collection_fetch_rejects_invalid_ids(self, ids):
+        collection = Collection.__new__(Collection)
+        collection._obj = MagicMock()
+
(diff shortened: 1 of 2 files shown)


HosniBelfeki:
Addressed Copilot review feedback: single empty-string IDs are now covered in fetch/delete validation tests, the non-string overload no longer overlaps with the single-string overload, and existing list inputs are reused without an unnecessary copy. I also updated this branch with the latest upstream main.

Cuiyus:
This PR is useful for improving Python-side error messages, but it does not appear necessary for native safety: pybind11 already validates conversion to std::vector<std::string>, while the C++ implementation safely handles missing keys.

PR #653, opened 2026-08-04
fix(python): normalize doc numpy vectors

HosniBelfeki:
Summary

This makes direct Doc(...) construction follow the documented vector normalization behavior.

Previously, Doc._from_tuple() converted numpy vector values to plain Python lists, but Doc.__init__ kept numpy.ndarray values unchanged. That could make repr(doc) fail JSON serialization and return the fallback error string for documents constructed directly with numpy vectors.

This routes both initialization paths through the same vector normalization helper and adds a regression test covering direct Doc(...) construction with a numpy vector.

Tests

- python -m ruff check python/zvec/model/doc.py python/tests/test_doc.py
- python -m ruff format --check python/zvec/model/doc.py python/tests/test_doc.py
- python -m py_compile python/zvec/model/doc.py python/tests/test_doc.py

Note: focused pytest against the local source tree is blocked in this workspace because zvec._zvec is not built locally.


HosniBelfeki:
Updated this branch with the latest upstream main after the linux-x64 run hit an unrelated native HNSW/RaBitQ test failure. The PR still only changes the Python Doc normalization path and its focused regression test. Local focused checks continue to pass: ruff check, ruff format --check, and py_compile for the changed files.

Cuiyus (approved):

PR #654, opened 2026-08-05
fix(kmeans): correct spherical centroid normalization

egolearner:
## Summary

Fix spherical K-means behavior for floating-point vectors by ensuring that
centroids are L2-normalized after each Lloyd iteration

## Motivation

The previous `Norm2` overloads selected the non-normalizing fallback for
floating-point contexts, leaving spherical centroids unnormalized.


iaojnh (approved):

richyreachy (approved):
LGTM

PR #655, opened 2026-08-05
fix(kmeans): correct spherical K-MC2 sampling weights

egolearner:
## Summary

Fix K-MC2 and AFK-MC2 initialization for spherical K-means when inner product is used as the underlying distance metric.

## Problem

K-MC2 initialization previously used raw negative inner-product scores as sampling weights. These scores can be negative and therefore are not valid sampling weights.

Spherical K-means is used with both normalized cosine vectors and unnormalized inner-product vectors, so the initialization logic must produce valid non-negative weights in both cases.

## Solution

Convert the raw negative inner-product score into the following spherical distance:

```text
||x|| - dot(x, c / ||c||)
```

This value is non-negative and reduces to cosine distance when both vectors are normalized.

The same conversion is applied consistently to:

- K-MC2 candidate scoring
- AFK-MC2 initial sampling probabilities
- Matrix-backed and cache-backed features
- Existing centroids supplied to spherical clustering

The behavior of non-spherical K-means is unchanged.

## Tests

Added coverage for:

- Non-negative spherical K-MC2 weights
- Normalized and unnormalized input vectors
- K-MC2 and AFK-MC2 initialization
- Matrix-backed and cache-backed features
- Existing spherical centroids
- Unchanged non-spherical behavior


egolearner:
测试结果
image

src/ailego/algorithm/kmeans.h
richyreachy: p_sum may be zero. 之前就有的问题，这样修复？：

double p_sum = std::accumulate(probs.begin(), probs.end(), 0.0);
const double uniform_prob = 1.0 / probs.size();

if (p_sum > 0.0 && std::isfinite(p_sum)) {
  for (auto &prob : probs) {
    prob =
        static_cast<float>((prob / p_sum + uniform_prob) * 0.5);
  }
} else {
  std::fill(probs.begin(), probs.end(),
            static_cast<float>(uniform_prob));
}

egolearner: done

richyreachy (approved):
LGTM

PR #513, opened 2026-06-22
feat(fts): add stemmer token filter based on Snowball 3.1.1

egolearner:
Implement a stemmer token filter for FTS that reduces words to their root form using the Snowball stemming library. Supports 34+ languages configurable via stemmer_lang in extra_params (defaults to english).

Changes:
- Integrate Snowball 3.1.1 as a thirdparty static library
- Add StemmerTokenFilter with thread_local stemmer cache (lock-free)
- Register 'stemmer' filter in TokenizerFactory
- Add unit tests and FtsColumnIndexer end-to-end tests

PR #661, opened 2026-08-06
fix(build): allow nmake for Snowball codegen on Windows

JalinWang:
fix #513 when MSYS2 does not present on Windows


egolearner (approved):
LGTM

PR #668, opened 2026-08-10
fix(build): refactor Snowball codegen build system to CMake for MSVC compatibility

JalinWang:
The Snowball codegen build process is defined by GUNMakefile, which is incompatible with `nmake`( bundled `make`-like tool with MSCV).  Translate the building script into cmake language, effectively removing the dependency of `make`.
Followup patch of #661.


egolearner (approved):
LGTM

PR #669, opened 2026-08-10
fix: compile MSVC sources as UTF-8

JalinWang:
## Summary

Configure MSVC to treat source and execution character sets as UTF-8.

## Motivation

On Windows systems using legacy active code pages such as CP936, MSVC can misinterpret UTF-8 source files, producing C4819 warnings and cascading syntax errors.

### Example failure

```text
uniform_int8_metric.cc(1): warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss
index_param.h(1): warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss.

index_param.h(68): error C4430: missing type specifier - int assumed.
index_param.h(75): error C2059: syntax error: '}'

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ if(MSVC)
             set(CMAKE_${_lang}_FLAGS_${_cfg} "${_flags}" CACHE STRING "" FORCE)
         endforeach()
     endforeach()
+    add_compile_options(/utf-8)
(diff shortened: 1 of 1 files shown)


JalinWang:
closed in favor of #666

PR #670, opened 2026-08-10
fix(ci): run checks when draft PR becomes ready

JalinWang:
Draft PRs skip lint and clang-tidy. Since ready_for_review is not a default pull_request activity, marking a PR ready does not trigger CI, leaving dependent builds skipped. Add ready_for_review while retaining the defaults.


egolearner (approved):
LGTM

PR #674, opened 2026-08-12
fix(db): remove orphaned segment directories during crash recovery

### LINKED ISSUE #673: Crash during Optimize can leak an orphaned segment dir and fail the next segment allocation after recovery
### Description

Since #614, Optimize moves the compacted output to its final numeric
directory and opens it in the lock-free phase, before the manifest is
persisted in the exclusive commit phase:

1. `allocate_segment_id()` returns N (in-memory fetch_add, not persisted)
2. `MoveDirectory(tmp -> <collection>/N)` — numeric dir now exists on disk
3. `Segment::Open(<collection>/N)` — heavy I/O (reads files, builds mmaps)
4. Optimize then waits for the exclusive schema lock; only the commit
   phase persists `next_segment_id = N+1` via version flush

If the process crashes between step 2 and the commit-phase flush,
recovery restores `segment_id_allocator_` from the old manifest (back
to N) and neither opens nor removes the unreferenced directory
`<collection>/N`. After recovery, the first operation that allocates
segment id N fails:

- an Insert that switches the writing segment hits
  `Segment::CreateAndOpen`'s existence check and returns
  "Segment create failed: segment path already exists" to the user;
- a schema DDL switching the writing segment fails the same way;
- the next Optimize fails at `MoveDirectory` (rename onto a non-empty
  directory, ENOTEMPTY).

Since `allocate_segment_id()` is fetch_add, the failure is one-shot
(a retry allocates N+1 and succeeds), but the orphaned directory — a
full compacted segment — is never referenced and never removed, i.e. a
permanent disk leak, and the first user operation after recovery can
spuriously fail.

The pre-#614 code had the same rename-before-flush window, but it was
microseconds wide (rename immediately followed by apply/flush in the
exclusive section). #614 materially widens it: the window now contains
`Segment::Open`'s I/O plus the wait for the exclusive schema lock
(unbounded if long readers hold the shared lock).

Expected: crash recovery leaves the collection in a state where
subsequent operations do not fail due to leftovers of an uncommitted
Optimize, and uncommitted output directories do not leak disk space.
I plan to submit a PR that removes segment directories not referenced
by the recovered manifest (numeric dirs absent from
`persisted_segment_metas()`, plus `*.tmp` residue) during `recovery()`,
where the collection file lock guarantees no concurrent Optimize.

### Steps to Reproduce

```python
1. Build a collection with enough persisted segments that Optimize produces a merged output (CompactTask)
2. Run Optimize() and kill the process after the tmp directory is renamed to its numeric path but before the commit phase flushes the manifest (e.g. inject a sleep/crash right after Segment::Open in the lock-free phase)
3. Reopen the collection (crash recovery), then insert until the writing segment switches, or run Optimize() again
4. The first operation that allocates the orphaned segment id fails; the orphaned numeric directory stays on disk permanently
```

### Logs / Stack Trace

```shell
Segment create failed: segment path already exists [<collection>/N]
move segment directory failed
```

### Operating System

macOS (Apple Silicon M5); applies to all platforms

Apple clang, CMake + Ninja (Homebrew), C++17; found by code inspection on main (post-#614), not platform-specific

### Additional Context

YongqiYin:
Fixes #673.

## Problem

Since #614, Optimize renames its compacted output to the final numeric
directory and opens it in the lock-free phase, before the commit phase
persists `next_segment_id`. If the process crashes inside that window,
recovery restores the id allocator from the old manifest and leaves the
unreferenced directory on disk:

- the directory leaks permanently;
- the first operation that re-allocates that id fails spuriously
  (`segment path already exists` on a writing-segment switch from
  Insert/DDL, ENOTEMPTY on the next Optimize's rename).

## Fix

`recovery()` now removes on-disk leftovers right after the manifest is
loaded, before opening segments:

- numeric segment directories absent from the recovered manifest
  (persisted set plus the writing segment);
- `<id>.tmp` compact outputs that were never renamed.

Safety:

- the exclusive collection file lock is already held, so no concurrent
  Optimize can be producing these directories;
- read-only opens hold only a shared lock and skip the cleanup;
- only names that round-trip through segment-id formatting (all digits,
  no leading zeros, fits uint32) are considered; anything else is left
  untouched;
- best-effort: a directory that cannot be removed is logged and skipped.

The Optimize three-phase locking model from #614 is unchanged.

## Test

New regression test `Feature_Recovery_Orphan_Segment_Dirs_Removed`:

- reads the manifest to learn the exact next segment id, then fakes the
  crash leftovers (colliding dir, far orphan dir, `*.tmp` residue) plus
  non-canonical guard dirs (`007`, `.tmp`, `5.tmpx`) that must survive;
- a read-only open removes nothing; a read-write open removes exactly
  the orphans and keeps every referenced segment directory;
- post-recovery `Optimize()` and an add-column DDL both succeed with all
  docs intact.

Fails on main at both the directory assertions and the post-reopen
`Optimize()`; passes with the fix. Full `collection_test` suite: 86/86.

Note: this makes a read-write `Collection::Open` delete directories under
the collection path that match canonical segment naming but are not
referenced by the manifest — a deliberate, logged self-heal for crash
leftovers.

### DIFF
--- a/src/db/collection.cc
+++ b/src/db/collection.cc
@@ -2050,7 +2050,8 @@ void CollectionImpl::cleanup_orphan_segment_dirs(const Version &version) {
   while (!ec && it != end) {
     std::error_code entry_ec;
     if (it->is_directory(entry_ec) && !entry_ec) {
-      std::string name = it->path().filename().u8string();
+      const std::string name =
+          ailego::FileHelper::PathToUtf8(it->path().filename());
 
       std::string stem = name;
       bool is_tmp = false;
@@ -2075,11 +2076,17 @@ void CollectionImpl::cleanup_orphan_segment_dirs(const Version &version) {
 
   for (const auto &name : orphan_names) {
     auto orphan_path = ailego::FileHelper::PathJoin(path_, name);
-    LOG_WARN("Removing segment directory left by an uncommitted operation: %s",
-             orphan_path.c_str());
-    if (!FileHelper::RemoveDirectory(orphan_path)) {
-      LOG_WARN("Failed to remove orphaned segment directory: %s",
-               orphan_path.c_str());
+    if (FileHelper::RemoveDirectory(orphan_path)) {
+      LOG_WARN(
+          "Recovery removed orphan segment directory not referenced by "
+          "manifest: path=%s",
+          orphan_path.c_str());
+    } else {
+      const auto error = ailego::FileHelper::GetLastErrorString();
+      LOG_WARN(
+          "Recovery failed to remove orphan segment directory not referenced "
+          "by manifest: path=%s, error=%s",
+          orphan_path.c_str(), error.c_str());
     }
   }
 }
--- a/src/db/collection.cc
+++ b/src/db/collection.cc
@@ -14,11 +14,14 @@
 
 #include <atomic>
 #include <cstdint>
+#include <filesystem>
+#include <limits>
 #include <memory>
 #include <mutex>
 #include <set>
 #include <shared_mutex>
 #include <string>
+#include <unordered_set>
 #include <variant>
 #include <vector>
 #include <ailego/io/file_lock.h>
@@ -148,6 +151,8 @@ class CollectionImpl : public Collection {
   Status create_idmap_and_delete_store();
 
   Status recover_idmap_and_delete_store();
+
+  void cleanup_orphan_segment_dirs(const Version &version);
 
   Status acquire_file_lock(bool create = false);
 
@@ -1934,6 +1939,17 @@ Status CollectionImpl::recovery() {
 
   auto segment_metas = v.persisted_segment_metas();
 
+  // A crash can leave segment directories on disk that the recovered
+  // manifest does not reference (e.g. Optimize renamed its output but did
+  // not reach the commit-phase flush). The recovered id allocator would
+  // collide with such a directory on the next allocation, so remove them
+  // now. The exclusive collection file lock is already held, hence nothing
+  // can be producing segment directories concurrently. Read-only opens
+  // hold only a shared lock and must not modify the collection.
+  if (!options_.read_only_) {
+    cleanup_orphan_segment_dirs(v);
+  }
+
   SegmentOptions seg_options;
   seg_options.read_only_ = true;
   seg_options.enable_mmap_ = options_.enable_mmap_;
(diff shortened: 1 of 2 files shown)


src/db/collection.cc
```diff
@@ -1990,6 +2002,88 @@ Status CollectionImpl::recover_idmap_and_delete_store() {
...
+  // matches the CleanupDirectory precedent.
+  std::vector<std::string> orphan_names;
+  std::error_code ec;
+  std::filesystem::directory_iterator it(
+      ailego::FileHelper::PathFromUtf8(path_), ec);
+  std::filesystem::directory_iterator end;
+  while (!ec && it != end) {
+    std::error_code entry_ec;
+    if (it->is_directory(entry_ec) && !entry_ec) {
+      std::string name = it->path().filename().u8string();
```
zhourrr: const std::string name =
    ailego::FileHelper::PathToUtf8(it->path().filename()); 改成这个吧，统一用 ailego filehelper 处理，以后对windows 等系统做特殊处理时也方便一点

src/db/collection.cc
```diff
@@ -1990,6 +2002,88 @@ Status CollectionImpl::recover_idmap_and_delete_store() {
...
+  }
+  if (ec) {
+    LOG_WARN("Failed to list collection directory for orphan cleanup: %s",
+             ec.message().c_str());
+    return;
+  }
+
+  for (const auto &name : orphan_names) {
+    auto orphan_path = ailego::FileHelper::PathJoin(path_, name);
+    LOG_WARN("Removing segment directory left by an uncommitted operation: %s",
```
zhourrr: if (FileHelper::RemoveDirectory(orphan_path)) {
  LOG_WARN(
      "Recovery removed orphan segment directory not referenced by manifest: "
      "path=%s",
      orphan_path.c_str());
} else {
  const auto error = ailego::FileHelper::GetLastErrorString();
  LOG_WARN(
      "Recovery failed to remove orphan segment directory not referenced by "
      "manifest: path=%s, error=%s",
      orphan_path.c_str(), error.c_str());
}

改成大概这样的逻辑吧。成功删除了就写 removed，删除失败了就写 failed to remove

YongqiYin: 已修改

src/db/collection.cc
YongqiYin: 已统一

zhourrr (approved):

PR #675, opened 2026-08-13
fix(cmake): clean stale submodule patch markers

JalinWang:
## Summary

- verify a submodule patch is still applied before trusting its marker
- remove stale markers and reapply missing patches during configure

## Why

`git submodule update` can restore tracked submodule files while leaving the untracked patch marker behind. A later configure then skips the patch and Linux RaBitQ compilation fails on missing patched APIs.


egolearner (approved):
LGTM

PR #694, opened 2026-08-21
fix(sqlengine): balance deep OR expression trees

Cuiyus:
## Summary

- consume deeply nested ANTLR OR parse trees iteratively
- rebuild large OR runs as balanced binary trees while preserving operand order, AND boundaries, and the historical shape of small expressions
- add a regression test for 4,096 non-foldable OR predicates and assert logarithmic tree depth

## Motivation

ANTLR represents a flat OR chain as a left-deep binary tree. The parser and downstream query-analysis paths recursively traverse that shape, so sufficiently large filters can overflow the worker stack. Balancing the OR tree reduces the logic depth from O(N) to O(log N).

## Validation

- CMake configuration completed successfully
- modified parser and regression-test translation units pass `clang++ -fsyntax-only`
- repository commit/push hooks passed clang-format, Ruff, gitleaks, conventional-commit, and branch-name checks

The full `query_info_test` binary was not built locally because this was a cold worktree build; CI is expected to run the complete target.


egolearner (approved):
LGTM

PR #708, opened 2026-08-27
feat(buffer): harden shared buffer pool core

### LINKED ISSUE #704: [Feature] Launch Buffer Storage for Memory-Bounded Query Caching
## Background

## Goals

## Scope

## Out of Scope

## User API

```python
import zvec

collection = zvec.open(
    "./items.zvec",
    option=zvec.CollectionOption(
        read_only=True,
        enable_mmap=False,
    ),
)
```

Release behavior:

## Implementation Phases

### Correctness

### Memory

### Performance

- QPS
- P95/P99
- Peak RSS
- Physical reads
- QPS/GiB

- Uniform unique
- 80/20 hotspot
- Zipf α=1.0
- 90/10 exact repeat

### Compatibility

## Definition of Done

iaojnh:
## Summary

This PR hardens the shared Buffer Pool core for future Buffer Storage integration with IVF, DiskANN, and other indexes. Index integrations, benchmarks, and user documentation are out of scope.

## Key changes

- Unify page buffers, metadata, and external caches under one process-wide memory budget.
- Harden concurrent page loading, pinning, eviction, and failure recovery.
- Add admission control, hot-page protection, and priority-aware CLOCK eviction.
- Support dirty-page tracking, batched writeback, failure recovery, and file extension.
- Add background reclamation with low-watermark control.
- Share Buffer Pool state across executables and DSOs.
- Consolidate io_uring support and reuse its reader types in DiskANN.
- Add statistics for memory usage, hits, eviction, admission, and writeback.

## Verification

- `vector_page_table_test`: 66/66 passed.
- Concurrent load-coalescing test: 2,000 repeated runs passed.
- CLOCK second-chance test: 1,000 repeated runs passed.
- clang-format 18 passed.
- MSVC and Linux Clang compatibility issues were fixed.


src/ailego/buffer/vector_page_table.cc
```diff
@@ -170,196 +626,976 @@ void VectorPageTable::release_block(block_id_t block_id) {
...
+                kVectorPageSize, std::chrono::milliseconds(100));
+        const auto wait_end = std::chrono::steady_clock::now();
+        writeback_wait_us_.fetch_add(
+            static_cast<uint64_t>(
+                std::chrono::duration_cast<std::chrono::microseconds>(
+                    wait_end - wait_start)
+                    .count()),
+            std::memory_order_relaxed);
+        const uint64_t now = writeback_pages_.load(std::memory_order_relaxed);
+        if (capacity_released || now != completed) {
```
richyreachy: 等待结束后立即再次尝试实际分配；成功就退出循环。如果仍然分配失败，只有后台完成了新的写回，才重置计数。

这里“收到容量可用的通知”会被直接当成进展，没有新写回、也分配不到内存时，就会耗尽重试预算并返回失败。

```
if (now != completed) {
  no_progress_waits = 0;
} else {
  ++no_progress_waits;
}
```

src/ailego/buffer/vector_page_table.cc
```diff
@@ -613,11 +2373,605 @@ void VecBufferPoolHandle::release_one(block_id_t block_id) {
...
+  }
+#endif
+
+  prefetch_pages_sync(first_page, page_count, priority);
+}
+
+void VecBufferPool::prefetch_pages_sync(block_id_t first_page,
+                                        size_t page_count, uint8_t priority) {
+  const size_t end_page = first_page + page_count;
+
```
richyreachy, line 2474: 这里可以为可写池做单独处理，可写池存在“加载期间内容被修改”的风险。

类似：
```
if (writable_) {
    for (size_t page_id = first_page; page_id < end_page; ++page_id) {
      char *buffer = acquire_buffer(page_id, 0, /*record_reuse=*/false);
      if (buffer == nullptr) {
        BlockEvictionQueue::get_instance().recycle();
        buffer = acquire_buffer(page_id, 0, /*record_reuse=*/false);
        if (buffer == nullptr) {
          break;
        }
      }
      page_table_.promote_evict_priority(page_id, priority);
      page_table_.release_block(page_id);
    }
    }
    return;
  }
```

src/ailego/buffer/vector_page_table.cc
iaojnh: 已修改。`wait_for_available()` 现在只负责唤醒重试，不再被视为实际进展；只有 `writeback_pages_` 增长时才会重置 `no_progress_waits`，避免容量短暂释放但分配仍失败时错误重置计数。

src/ailego/buffer/vector_page_table.cc
iaojnh: 已按建议修改。可写池的同步预取现在逐页通过 `acquire_buffer()` 加载，复用正常的 single-flight 状态机，避免并发写入、写回和驱逐后装入旧的磁盘内容；只读池仍保留批量预取路径。

richyreachy (approved):

PR #715, opened 2026-08-31
fix(db): SIGSEGV when fetch/query race a segment switch; unblock reader concurrency

### LINKED ISSUE #714: [Bug]: SIGSEGV under concurrent fetch + insert when the writer crosses a segment switch
### Description

One writer thread running `insert()` concurrently with reader threads running `fetch()` / `query()` / `group_by_query()` crashes the process with SIGSEGV on Linux when the writer crosses a segment-switch boundary. Reproduced **3/3 on `main` .

**Root cause.** Readers snapshot the segment list via `get_all_segments()` with **no lock**, while writers mutate that state under exclusive `write_mtx_` ,`doc_ids_` grows per insert , and `writing_segment_` is reassigned per switch. Worse, `dump()`/`flush()` tear down the old segment's in-memory store (`memory_store_.reset()`) **without taking `seg_mtx_` at all** , so a reader mid-`Fetch` — even holding `seg_mtx_` — reads a store that is being destroyed.

gdb captures both threads in the same instant (full stacks below):

- **Reader**: `fetch` → `SegmentImpl::Fetch` → `MemForwardStore::convertToTable` → `arrow::Table::FromRecordBatches` → SIGSEGV
- **Writer**: `write_impl` → `switch_to_new_segment_for_writing` → `dump()` → `flush()` → `LocalWalFile::remove()`

On macOS the same race does not crash; a probe (150µs widened reader window) measured **99.7%** of `doc_ids_` push_backs and **99.4%** of segment-switch reassignments executing while a reader was inside `get_all_segments()`. The unsynchronized shape dates to the initial commit.

**Why rarely hit**: only the doc-count-threshold switch is exposed — the other seven switch call sites (optimize, DDL, iterator creation) hold the schema lock exclusively and drain in-flight readers first (verified: 285 optimize-triggered switches, zero crashes). But the threshold counts **cumulative inserts, not collection size** (`doc_ids_` is append-only), so any long-running writer eventually crosses it.

### Steps to Reproduce

```python
1. Build zvec on Linux x86_64 (glibc), `CMAKE_BUILD_TYPE=Release`.
2. Collection with a vector field; `max_doc_count_per_segment` = 4,000 (schema-validated: `validate()` accepts ≥ 1,000) and 8 MB forward buffer — this compresses the switch interval from ~10M inserts to seconds without changing the code path.
3. Preload ~3,000 docs.
4. One writer thread: `insert()` batches of 100 in a tight loop.
5. Four reader threads: `fetch()` preloaded docs by pk in a tight loop.
6. SIGSEGV within seconds.
```

### Logs / Stack Trace

```shell
Thread (reader) — SIGSEGV:
#0  arrow::Field::ToString() const
#1  arrow::Schema::ToString() const
#2  arrow::Table::FromRecordBatches(...)
#3  zvec::MemForwardStore::convertToTable(...)
#4  zvec::MemForwardStore::fetch(...)
#5  zvec::SegmentImpl::fetch_normal(...)
#6  zvec::SegmentImpl::Fetch(unsigned long, ...)
#7  zvec::CollectionImpl::fetch(...)

Concurrent writer thread (same instant):
#2 std::filesystem::remove(...)
#3 zvec::ailego::FileHelper::DeleteFile(char const*)
#4 zvec::LocalWalFile::remove()
#5 zvec::SegmentImpl::flush()
#6 zvec::SegmentImpl::dump()
#7 zvec::CollectionImpl::switch_to_new_segment_for_writing(...)
```

### Operating System

Linux x86_64 (Alibaba Cloud ECS, 64-core Xeon Platinum 8369B, glibc 2.32, GCC 10.2.1). Not observed to crash on macOS arm64 under the same load (race still active there; see Description).

 zvec `main` @ `3d479a1` (also reproduced at `733903c`). CMake + Ninja, Release, C++17.

### Additional Context

YongqiYin:
Closes: #714

## What

Both read entry points crash with SIGSEGV on Linux when a writer crosses a
segment switch (details in #714). `fetch()` faults inside the in-memory table
rebuild; `query()` faults on the same store, reached through the planner's
fan-out (`VectorRecallNode::collect_batch → SegmentImpl::fetch → fetch_normal
→ convertToTable`). Two races and two read-path changes:

1. **Segment list** — `get_all_segments()` takes `write_mtx_` shared
   (`collection.cc`), as `stats()` already does; `_unsafe` variant added for
   callers already holding the lock. Probe on `main`: 99.7% of `doc_ids_`
   push_backs executed while a reader was inside this function.
2. **Segment teardown** — `finish_memory_components()` /
   `init_memory_components()` take `seg_col_mtx_` exclusively, and the two
   combined-vector-indexer accessors take it shared; `dump()` takes `seg_mtx_`
   exclusively. Readers used to reach `memory_store_` / `persist_stores_` /
   the indexer maps while `flush()` was republishing them under a *different*
   lock, so a reader landing between `memory_store_->close()` and
   `persist_stores_.push_back()` found neither and dropped that block's rows
   with no error and no log. Locking the rebuild covers all three `flush()`
   entry points (buffer-full insert, segment switch, close).
3. **Point-read fast path** — `MemForwardStore::convertToTable()` materializes
   only the batch holding the requested rows instead of rebuilding the whole
   in-memory store: O(store) → O(1), and it no longer walks the stack the
   crash faults in.
4. **Reader locks** — `seg_mtx_` and `cache_mtx_` become `shared_mutex`;
   read-only accessors take them shared (verified member by member).
   `close()` now takes `cache_mtx_` exclusively, with `flush_locked()`
   extracted so it does not lock twice.

Also from review: the three `convertToTable` helpers moved to snake_case, and
`tests/db/concurrent_fetch_test.cc` was renamed `concurrent_read_test.cc` now
that it covers both read paths.

## Test plan

`tests/db/concurrent_read_test.cc`, one writer crossing frequent segment
switches throughout:

- **fetch**: 4/8 readers comparing every field of every fetched doc against
  the deterministic generator output. **3/3 SIGSEGV on unfixed `main`**, passes
  with the fix (2.4M field comparisons on Linux, 1.6M on macOS, 0 errors).
- **query**: 4 queriers running KNN with an output field, asserting no errors,
  no empty results, no missing field values, and that each value matches the
  pk it came back with. **3/3 SIGSEGV on unfixed `main`**, passes with the fix
  (19.5k queries on Linux, 16.5k on macOS, 0 errors).

Two settings are load-bearing: `max_doc_count_per_segment` = 4,000
(schema-validated ≥1,000) so switches happen inside the run window, and the
writer is capped at 50k docs and **asserted to be alive** (`writer_errors` plus
a minimum `docs_written`) — otherwise a starved writer silently turns the run
into a no-writer one and the test passes without exercising the crash path.

Full suites: 222 tests on macOS arm64, 224 on Linux x86_64.

## Measured impact (64-core Linux; same-binary repeat variance <4%)

| metric | before | after |
|---|---|---|
| point-read latency (500→4,000 docs in the in-memory store) | 472→1,982µs, grows with doc count (O(N)) | flat ~180µs at any doc count (O(1)) |
| fetch throughput 1→8 threads, **no writer** | flat (no reader-reader parallelism) | **5.8x scaling** (5.6k→32k reads/s) |
| fetch throughput 1→8 threads, **with writer** | crashes | **6.4x scaling** (8.7k→56k reads/s) |
| query throughput 1→8 threads, no writer | **1.15x** (597→685 q/s) — materialization serialized under the exclusive lock | **3.4x** (600→2,038 q/s) |
| writer throughput as readers 0→8 | starves (290 docs/s, macOS) / crashes (Linux) | flat (−2.4%) |

## Known trade-offs 
- **Reads now wait for the in-flight write batch.** Before this PR, readers
  fetched the segment list with no lock at all — that unlocked read is exactly
  the race being fixed. After it, `get_all_segments()` takes `write_mtx_`
  shared, so a reader waits for the current `write_impl` batch. This is the
  only newly-blocked path at the collection layer: `create_index` /
  `add_column` already blocked readers for the whole DDL task (they hold the
  schema lock exclusively), and per-insert exclusion already existed at the
  segment layer.
- **Reads now wait for a segment teardown or rebuild.** Item 2 blocks readers
  for the duration of `dump()` (avg 16ms, max 24ms, 26 occurrences over an
  8s window in the regression test) and of a memory-component rebuild.
  Previously they were not blocked — they read state that was being torn down.
  Measured cost is inside the ±1% run-to-run noise on every workload above.
- **Writers can be passed by concurrent readers.** With `std::shared_mutex`, a
  writer waiting for a segment lock must wait out the readers currently inside
  it. The wait stays bounded: new readers cannot arrive (the writer holds the
  collection's exclusive `write_mtx_`) and each in-flight read is now an O(1)
  lookup. Writer throughput stays flat as readers go 0→8 (−2.4%).


src/db/index/segment/segment.cc
```diff
@@ -415,7 +415,13 @@ class SegmentImpl : public Segment,
 
   bool sealed_{false};
 
-  mutable std::mutex seg_mtx_;
+  // Guards the segment state rewritten by Insert/Update/Upsert/Delete and the
+  // buffer-full flush() they trigger under this lock: doc_ids_, memory_store_,
+  // persist_stores_, the vector indexer maps and the block metadata. Read-only
+  // accessors (Fetch, get_global_doc_id) take it shared. NOT covered: the
+  // segment-switch dump()/flush(), which rewrites the same state without
+  // seg_mtx_ and is serialized only by the collection's write_mtx_.
```
egolearner, line 423: Collection中get_segments时虽然加了write_mtx_读锁，但是得到segment列表后就释放了锁。后续在调用Segment::fetch时，有可能其他地方调用了SegmentImpl::flush或dump，修改了memory_store_导致有并发问题。

src/db/index/storage/memory_forward_store.cc
```diff
@@ -93,27 +93,33 @@ bool MemForwardStore::validate(const std::vector<std::string> &columns) const {
 
 // Notice: This function just convert the docs to arrow::ArrayBuilder, not clean
 // the cache_.
+arrow::Status MemForwardStore::appendDocToBuilder(
```
egolearner, line 96: 新增代码的函数使用`snake_case`风格吧，存量代码的风格问题可以后续再另行修复

src/db/index/segment/segment.cc
```diff
@@ -415,7 +415,13 @@ class SegmentImpl : public Segment,
...
+  mutable std::shared_mutex seg_mtx_;
```
egolearner, line 424: `seg_mtx_` 和 `seg_col_mtx_` 保护的状态存在重叠，包括 `memory_store_`、`persist_stores_` 和 block metadata。使用两把锁容易导致 `flush()` 和 `scan()` 在不同锁下访问同一状态。建议合并为一把 `shared_mutex`：读操作持共享锁，写入、flush、dump 和 DDL 持独占锁。同时需要引入 `_unsafe` 或内部无锁 helper，避免 `Fetch() -> fetch()` 这类调用链重复获取不可重入的 mutex。

src/db/index/storage/memory_forward_store.cc
chinaux, line 406: 这里也需要加个锁  std::lock_guard<std::shared_mutex> lock(cache_mtx_); 或者更干净做法（改动大一点）：抽一个不加锁的 flush_locked()，flush() = 加锁 + flush_locked()，close() 则一次性持独占锁后调 flush_locked() + 清理。

tests/db/concurrent_fetch_test.cc
```diff
@@ -0,0 +1,215 @@
...
+    zvec::ailego::MemoryLimitPool::get_instance().init(4 * 1024ll * 1024ll *
+                                                       1024ll);
+    ailego::FileHelper::RemoveDirectory(kPath);
+  }
+  void TearDown() override {
+    ailego::FileHelper::RemoveDirectory(kPath);
+  }
+};
+
+TEST_F(ConcurrentFetchTest, FetchReturnsCorrectContentUnderConcurrency) {
```
chinaux: 只压了 fetch()，建议再加一个 并发 query 的测试case

src/db/index/storage/memory_forward_store.cc
YongqiYin: 第一种因为 flush自己也取排他cache_mtx，会死锁。已采纳第二种做法

src/db/index/storage/memory_forward_store.cc
YongqiYin: done

tests/db/concurrent_fetch_test.cc
YongqiYin: 已加，测试期间writer跨10余次段切换query全通过

src/db/index/segment/segment.cc
YongqiYin: 已修，重写 memory_store_ / persist_stores_ 的finish_memory_components()只被 flush() 调用，现在取排他 seg_col_mtx_，所以 flush 的三条触发路径（internal_insert 的 buffer-full、dump、close）都被覆盖；dump() 另外取排他 seg_mtx_

src/db/index/segment/segment.cc
YongqiYin: 已按建议合并为一把 shared_mutex：读持共享锁，写入/flush/dump/DDL/内存组件重建持独占锁；memory_store_、persist_stores_、block metadata 现在都在同一把锁下，消除了两把锁分别访问同一状态的问题。递归重入用无锁 helper 解决：finish/init_memory_components 改为调用方持锁的无锁内核，fetch(columns,int) 拆出 fetch_exec_unsafe 供 Fetch(doc) 直调避免 Fetch()→fetch() 重复加锁。
测试吞吐无回退，写者无饥饿。

YongqiYin:
评审期间上游合入了 #722（synchronize reads of writing segment）和 #726（synchronize flat cache and
collection block switching），修的是本 PR 同一批竞态，但方向相反：

- **上游**：query/fetch 全程持 shared `write_mtx_`，相当于查询期间冻结整个集合。语义简单、一致性强，
  但读取进行期间写者无法推进。
- **本 PR**：仅在取段列表时短暂持 shared `write_mtx_`，之后由段级 shared 锁保证段内容一致，读写并发。

把两套方案在同一台机器上做了对照实测。

### 测试条件

`tests/db/concurrent_read_test.cc`，（Linux，64 核），每组 3 次，8 秒窗口：

- **写侧**：1 个线程按 100 条/批持续 insert，上限 50,000 条；`max_doc_count_per_segment=4000`，
  运行期间会反复跨段切换
- **读侧**：4 个线程做点查（逐字段校验内容），或 4 个线程做 KNN topk=10
- 读写同时进行，下表三个指标取自同一次运行

### 结果

| 指标 | 上游锁方案 | 本 PR | 倍数 |
|---|---|---|---|
| fetch 吞吐（4 读者 + 1 写者） | 5,135–5,352/s | 88,564–89,380/s | ~17x |
| query 吞吐（4 查询者 + 1 写者） | 1,672–1,718/s | 2,441–2,455/s | ~1.4x |
| **写者吞吐（与上面读负载并发）** | **12 docs/s**——8 秒内只完成 1 批 100 条 | **6,250 docs/s**——写满 50,000 条上限 | **~520x** |

写者停滞的原因是读者饥饿：glibc 的 rwlock 读者优先，读者全程持 shared `write_mtx_` 时，持续的读流量
会让排他写者长时间拿不到锁。macOS 的 libc++ 写者优先，因此只在 Linux 复现。该现象已用**纯上游代码**
（不含本 PR 任何改动）独立复现 3/3。

### 正确性对齐

#722 移除了 `collection_test.cc` 中对 `fetch table failed` 瞬态的容忍，断言更严格。本 PR 在该测试下
Linux 5/5、macOS 8/8 通过；新增的高压并发测试约 5.7 万次 query 也是 `tolerated=0`——段锁合并后
`internal_insert` 的元数据发布与读者严格互斥，顺带消除了这个瞬态。

### 合并的处理

保留本 PR 的锁设计，并吸收上游的覆盖面：补上 `get_vector_indexer()` /
`get_quant_vector_indexer()` 的 shared 锁。

src/db/index/segment/segment.h
```diff
@@ -65,6 +65,20 @@ class Segment {
 
   virtual SegmentMeta::Ptr meta() const = 0;
 
+  // Doc-id bounds for routing. The default reads the meta unlocked; SegmentImpl
+  // overrides it to take the shared segment lock, so concurrent callers must
+  // not read min/max_doc_id() off meta() directly.
+  virtual void doc_id_range(uint64_t *min_id, uint64_t *max_id) const {
```
egolearner: SegmentImpl会重写，这里提供一个不加锁的基类实现的作用是什么？

src/db/collection.cc
```diff
@@ -1771,13 +1777,8 @@ Status CollectionImpl::delete_by_filter(const std::string &filter) {
...
   query.include_doc_id_ = true;
 
-  // A query must see a stable writing segment. Writers publish the forward
-  // store, scalar/vector indexes and segment metadata in several steps while
-  // holding write_mtx_ exclusively.
-  auto ret = [&]() {
-    std::shared_lock<std::shared_mutex> write_lock(write_mtx_);
-    return sql_engine_->execute(schema_, std::move(query), get_all_segments());
-  }();
+  auto ret =
```
egolearner: write_mtx读锁去掉了会影响一致性？

egolearner: 其他的去掉不一定有问题，但是delete_by_filter还是需要一个强一致的保证，建议保留

YongqiYin: 同意，已恢复读锁

src/db/index/segment/segment.h
YongqiYin: 没用，已改为纯虚，MockSegment中补了对应显示实现

src/db/collection.cc
```diff
@@ -209,8 +209,14 @@ class CollectionImpl : public Collection {
   Result<std::unique_ptr<DocIterator::Impl>> prepare_iterate(
       const IteratorOptions &options);
 
+  //! Collects the segment list under a shared write_mtx_, because
+  //! writing_segment_ and the doc_ids_ that doc_count() reads are mutated by
+  //! Insert under the exclusive write_mtx_.
   std::vector<Segment::Ptr> get_all_segments() const;
 
+  //! Same as get_all_segments(), for callers that already hold write_mtx_.
+  std::vector<Segment::Ptr> get_all_segments_unsafe() const;
```
egolearner, line 217: private？

src/db/index/segment/segment.cc
```diff
@@ -1352,9 +1376,9 @@ Doc::Ptr SegmentImpl::Fetch(
 
 CombinedVectorColumnIndexer::Ptr SegmentImpl::get_combined_vector_indexer(
     const std::string &field_name) const {
-  // Copy indexers and their block metadata together; flush/init can run
-  // concurrently with query planning.
-  std::shared_lock<std::shared_mutex> lock(seg_col_mtx_);
+  // Shared: finish_memory_components() migrates entries between these maps,
+  // so an unlocked read can miss an indexer or count it twice.
+  std::shared_lock<std::shared_mutex> lock(seg_mtx_);
```
egolearner, line 1375: [P1] The reader lock does not cover all writers of the segment metadata. In CollectionImpl::switch_to_new_segment_for_writing(), remove_writing_forward_block() (collection.cc:1730) runs after dump() has released seg_mtx_. A concurrent query can therefore pass has_writing_forward_block() in the CombinedVectorColumnIndexer constructor and race with the optional reset before value() or the BlockMeta copy. This is a data race with potential exceptions/crashes, not merely a weakly consistent read. Could we move this metadata removal into a Segment operation protected by the exclusive seg_mtx_?

YongqiYin: 已修，采纳建议，新增Segment::remove_writing_forward_block()，在SegmentImpl里取 seg_mtx_ 排他后再改 meta。

egolearner (approved):
LGTM

PR #718, opened 2026-09-02
feat(turbo): add SSE distance kernels

richyreachy:
**Summary**
Add SSE distance implementations for Turbo on x86 platforms.
This PR adds single-vector and batch kernels for:

- FP32
- FP16
- Record-quantized INT8
- Packed record-quantized INT4

Supported metrics:

- Squared Euclidean
- Cosine
- Inner product

The new kernels are registered between AVX2 and scalar implementations in Turbo’s runtime dispatch table. FP16 conversion uses SSE2 without requiring F16C and correctly handles zero, subnormal, infinity, and NaN encodings.
The SSE build configuration now targets the same SSE2 baseline used by runtime feature detection, preventing compiler auto-vectorization from introducing newer instructions. The existing SSE FHT implementation also uses the shared portable SSE2 feature guard, including support for MSVC x64.


src/turbo/turbo.cc
```diff
@@ -91,6 +91,10 @@
 #include "scalar/record_quantized_int8/inner_product.h"
 #include "scalar/record_quantized_int8/squared_euclidean.h"
 #include "scalar/rotate/fht/fht.h"
+#include "sse/fp16/distance.h"
```
JalinWang: Rename this to sse2 to be more precise?

src/turbo/CMakeLists.txt
```diff
@@ -4,6 +4,18 @@ include(${PROJECT_ROOT_DIR}/cmake/option.cmake)
 if(NOT ANDROID AND AUTO_DETECT_ARCH)
     if (HOST_ARCH MATCHES "^(x86|x64)$")
         setup_compiler_march_for_x86(TURBO_MARCH_FLAG_SSE TURBO_MARCH_FLAG_AVX2 TURBO_MARCH_FLAG_AVX512 TURBO_MARCH_FLAG_AVX512FP16)
+        # kSSE is gated by SSE2 at runtime. Keep these translation units at
```
JalinWang: Remane the CpuArch flag to kSSE2 to be more precise, as well as the folder/namespace? We might have to deal with SSE4.2 CRC instruction somewhere in turbo.

Image

richyreachy: changed SSE->SSE2 already. 

That's right. CRC is another issue, which uses attribute target to do runtime dispatch currently. we can take it into turbo later.

JalinWang (dismissed):

JalinWang (approved):

PR #733, opened 2026-09-08
perf(ivf): reduce peak memory usage during index build and merge

iaojnh:
## Summary

Reduce peak memory usage during IVF index construction and merge by eliminating redundant vector copies, bounding temporary buffers, and releasing completed build state earlier.

The optimization supports both unquantized FP32 input and eligible FP32-to-FP16 conversion paths without changing the Python API.

## Implementation

- Introduce a provider-backed, multipass merged holder that reads source vectors directly. Filtering is evaluated once, with consistent document-ID remapping across traversals.
- Use ordinal access in eligible IVF builds to retain lightweight source references instead of a full vector copy, retrieving vectors during dump.
- Support ordinal access through the FP16 conversion holder. Each reader converts one source vector at a time and maintains independent buffers, avoiding a full-dataset FP16 copy.
- Populate the KMeans training matrix directly from the holder on supported FP32/L2 paths, removing the intermediate training-feature copy.
- Bound outstanding label-generation vector payloads to a 4 MiB budget, allowing a single oversized vector to make progress.
- Release the builder after a successful dump, before opening the resulting index, and release remaining build inputs after finalization succeeds.

Sources and transformations without ordinal-access support retain the materialized fallback. The KMeans training matrix and source-reader metadata are still required; the buffer budget is not a total-memory limit.

## Correctness

- Preserve source ownership for the lifetime of readers.
- Handle reused or unaligned provider buffers safely.
- Propagate provider-read, conversion, and dump failures.
- Preserve retryable state after partial failures without repeating completed build stages.
- Support reader reset, repeated dumps, and consistent fetch/search behavior after reopening.

## Validation

All 111 tests across five related Linux ARM64 test targets passed:

- `ivf_builder_test`
- `ivf_searcher_test`
- `index_interface_test`
- `merged_provider_index_holder_test`
- `half_float_reformer_test`

The three new FP16 ordinal-access tests each passed 20 repeated runs. clang-format 18.1.8 and whitespace checks also passed.

## Peak RSS

Benchmarks use four Flat sources containing a total of 1,000,000 × 128-dimensional FP32/L2 vectors on Linux ARM64. The main baseline is `930d243b`.

Each result is the median of three independent process runs. Peak RSS is measured using `wait4` / `ru_maxrss`.

| Configuration | main | Optimized | Reduction |
|---|---:|---:|---:|
| FP32, no sampling, nlist=1 | 2680.2 MiB | 1077.3 MiB | 59.8% |
| FP32, 10,000 training samples, nlist=64 | 2197.6 MiB | 614.1 MiB | 72.1% |
| FP16 target, 10,000 training samples, nlist=64 | 1953.5 MiB | 615.0 MiB | 68.5% |

All measured builds passed document-count and boundary-vector checks. Full source-vector verification also passed, with no OOM events.

On-demand FP16 conversion trades additional source reads and conversion work for lower memory usage. Actual performance depends on the workload and build configuration.

These measurements cover C++ IVF merge, not end-to-end Python `Collection.optimize()`. The nlist=1 case exercises full-data training.


src/core/quantizer/half_float_converter.cc
```diff
@@ -34,8 +36,61 @@ turbo::ConvertFunc ResolveFp16ConvertFunc() {
...
+      const void *source_data = nullptr;
+      int ret = reader_->read(ordinal, &source_key, &source_data);
+      if (ret != 0) {
+        return ret;
+      }
+      if (!source_data) {
+        return IndexError_Runtime;
+      }
+
+      // Providers may return reused, unaligned storage. Keep only one aligned
```
JalinWang: Fp16内核似乎没有要求对齐？后面transform_record也没对齐

iaojnh: 确认了，FP16 转换不要求 SIMD 对齐，原来的无条件拷贝确实多余。已改为正常输入直接转换，仅在地址不满足 `alignof(float)` 时才拷贝，保证 scalar/tail 路径的读取安全。两种情况的测试也已补充并通过。

JalinWang (approved):

Issue #309, opened 2026-04-03
🌟 Roadmap: v0.5.0 and Beyond


zhourrr:
This issue tracks the high-level roadmap and upcoming features for Zvec. Our focus for the next major iterations is on **storage scalability**, **mobile ecosystem maturity**, and **deeper integration with the AI landscape**.

### 📱 Platform & Ecosystem Expansion
* [x] **Full Mobile Support:** Complete iOS compilation support and provide idiomatic SDKs for both Android and iOS with comprehensive examples.
* [x] **Broader Language Bindings:** Expand SDK capabilities to include Go, Rust, and Dart (Flutter).
* [ ] **Framework Integration:** Broaden support for RAG/Memory ecosystems and extend AI extensions to include more embedding models and ETL pipelines.
* [ ] **Database Extensions:** Explore integration with embedded databases such as **SQLite** and **DuckDB**.

### 💾 Storage & Resource Management
* [ ] **Disk-Based Indexing:** Implementation of DiskANN-inspired algorithms to support datasets exceeding RAM capacity.
* [ ] **RabitQ Enhancements:** Further optimization of quantization logic for better accuracy-efficiency trade-offs.
* [ ] **Memory Governance:** Introduce strict memory limits and constraints during the `CreateIndex` process to ensure system stability in resource-constrained environments.

### ⚙️ Feature & Performance Engineering
* [ ] **Full-Text Search:** Introduce keyword search capabilities alongside vector retrieval.
* [ ] **Optimization Pipeline:** Refine the `Optimize` workflow and improve write throughput for high-concurrency scenarios.
* [ ] **ARM-Specific Tuning:** Dedicated performance profiling and optimization for ARM-based architectures (mobile and Graviton).
* [ ] **Benchmarking:** Consistent participation in **ANN-Benchmarks** to track performance parity with industry-leading libraries.

### 🛡️ Stability & Quality Assurance
* [ ] **Testing Coverage:** Continuous expansion of API integration tests and code coverage.
* [ ] **Regression Efficiency:** Enhance CI/CD pipelines for faster and more reliable regression testing.

### 💬 We want your feedback\!
This roadmap is a living document. Please head over to our [Discussion Thread](https://github.com/alibaba/zvec/discussions/310) to suggest new features.

PR #737, opened 2026-09-09
feat(thirdparty): add official sqlite submodule

ilike314272:
Added sqlite third party library, supported on Windows, Apple, and Linux. Aligns with roadmap #309


zhourrr:
Thanks for the contribution. Our roadmap is to provide Zvec as a SQLite
extension, rather than integrate SQLite into Zvec.

Could you clarify how this SQLite dependency will support that goal? Since no
extension target currently uses it, it may be clearer to introduce the
dependency together with the extension implementation and tests.

ilike314272:
> Thanks for the contribution. Our roadmap is to provide Zvec as a SQLite extension, rather than integrate SQLite into Zvec.
> 
> Could you clarify how this SQLite dependency will support that goal? Since no extension target currently uses it, it may be clearer to introduce the dependency together with the extension implementation and tests.

Thank you for your guidance. I misread the roadmap. Give me some time to come back with a sqlite extension that uses the `sqlite3_module` interface. I will close this PR for cleanliness.

PR #738, opened 2026-09-09
perf(vamana): optimize search state and fix medoid selection

luoxiaojian:
- Inline BlockHeap implementation.
- Use bitmap as the default Vamana visit filter.
- Exclude the UniformUint8 norm tail from medoid calculation.
- Include regression tests and build configuration updates.


richyreachy (approved):
LGTM

PR #742, opened 2026-09-10
feat(buffer): add buffer-backed storage and HNSW integration

### LINKED ISSUE #704: [Feature] Launch Buffer Storage for Memory-Bounded Query Caching

iaojnh:
## Summary

- Extend `IndexStorage` with lifetime-safe borrowed, batched, immutable, and scatter reads.
- Add read-only `BufferReadStorage` and complete writable `BufferStorage` support, including concurrent writes, flush, resize, and recovery handling.
- Integrate HNSW build and search with Buffer Storage using batched page reads, pinned vectors, and batched graph updates.
- Adapt Flat and Vamana to the new storage lifetime model.
- Optimize `BlockHeap` and harden `VisitFilter` lifecycle and error handling.
- Add coverage for Buffer Storage reads/writes, memory-block lifetime, HNSW integration, and search regressions.

## Testing

- Built `zvec_core` and `local_builder`.
- Passed `hnsw_streamer_buffer_test`.
- Passed `block_heap_test`.
- Passed `vamana_fast_search_test`.
- Passed clang-format and clang-tidy checks.

IVF, DiskANN, RocksDB memory budgeting, public API exposure, documentation, and benchmarks are intentionally left for follow-up PRs.

### DIFF
--- a/src/core/algorithm/hnsw/hnsw_algorithm.cc
+++ b/src/core/algorithm/hnsw/hnsw_algorithm.cc
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 #include "hnsw_algorithm.h"
-#include <type_traits>
 
 namespace zvec {
 namespace core {
@@ -183,12 +182,12 @@ void HnswAlgorithm<EntityType>::add_neighbors(node_id_t id, level_t level,
 
   HnswDistCalculator &dc = ctx->dist_calculator();
 
-  update_neighbors(dc, id, level, topk_heap);
+  update_neighbors(dc, id, level, topk_heap, ctx);
 
   // reverse update neighbors
   for (size_t i = 0; i < topk_heap.size(); ++i) {
     reverse_update_neighbors(dc, topk_heap[i].first, level, id,
-                             topk_heap[i].second, ctx->update_heap());
+                             topk_heap[i].second, ctx->update_heap(), ctx);
   }
 
   return;
@@ -200,11 +199,13 @@ void HnswAlgorithm<EntityType>::add_neighbors(node_id_t id, level_t level,
 // Two specialized inner loops, dispatched from search_neighbors():
 //
 //   fast_search_neighbors:       mmap/contiguous with direct vector pointers.
-//                                Uses BlockHeap (AVX2) or LinearPool (scalar)
-//                                for visited tracking and top-k maintenance.
+//   fast_search_neighbors_buffer: BufferStorage with page-backed MemoryBlocks.
+//                                Both use BlockHeap (AVX2) or LinearPool
+//                                (scalar) for visited tracking and top-k
+//                                maintenance.
 //   dual_heap_search_neighbors:  CandidateHeap + TopkHeap + VisitFilter.
 //                                Used for add_node (use_pool=false), filtered
-//                                search, upper levels, and BufferPool fallback.
+//                                search and upper levels for every backend.
 // ============================================================================
 
 // mmap/contiguous variant: resolve vectors via get_vector_ptr and use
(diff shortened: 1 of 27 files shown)


src/core/utility/buffer_storage.cc
```diff
@@ -449,19 +905,120 @@ class BufferStorage : public IndexStorage {
...
+          force_bypass ? owner_->buffer_pool_handle_->read_range_bypass(
+                             abs_offset, len, tmp)
+          : immutable ? owner_->buffer_pool_handle_->read_range_immutable(
+                            abs_offset, len, tmp)
+                      : owner_->read_range(abs_offset, len, tmp);
+      if (!read_ok && immutable && !force_bypass) {
+        // A cross-page immutable read can lose a capacity race after choosing
+        // the cache path. The bytes are immutable, so retry directly instead
+        // of failing the HNSW expansion.
+        read_ok = owner_->buffer_pool_handle_->read_range_bypass(abs_offset,
```
richyreachy, line 1017: 跨页读取因缓存容量不足而失败，会从磁盘重新读取。但缓存中可能存在尚未刷盘的数据，“immutable”不等于“persisted”。应保留缓存中的有效片段，或先刷盘再读取。

src/core/algorithm/hnsw/hnsw_algorithm.cc
```diff
@@ -644,24 +723,72 @@ void HnswAlgorithm<EntityType>::update_neighbors(HnswDistCalculator &dc,
...
+    auto &prune_blocks = ctx->prune_blocks();
+    auto &selected_indices = ctx->prune_selected_indices();
+
+    prune_ids.clear();
+    prune_ids.reserve(topk_heap.size());
+    for (const auto &candidate : topk_heap) {
+      prune_ids.emplace_back(candidate.first);
+    }
+    prune_blocks.clear();
+    if (ailego_unlikely(read_entity.get_vector_for_prune(prune_ids.data(),
```
richyreachy: 新的优化路径直接读取索引中的向量，但距离计算器可能使用外部 provider 对应的度量。

tools/core/local_builder.cc
```diff
@@ -561,8 +557,7 @@ int do_build_by_streamer(IndexStreamer::Pointer &streamer,
   }
 
   IndexQueryMeta qmeta(holder->data_type(), holder->dimension());
-  const size_t keep_docs = holder->count();
-  const size_t end_cursor = holder->end_cursor();
+  uint32_t keep_docs = holder->count() - holder->start_cursor();
```
richyreachy: 这里是不是做了重复扣除？

src/core/utility/buffer_storage.cc
```diff
@@ -243,105 +213,535 @@ class BufferStorage : public IndexStorage {
...
+              state.abs_offset % ailego::kVectorPageSize;
+          owner_->buffer_pool_handle_->acquire_one(
+              scratch.unique_page_ids[unique_index]);
+          state.request->block->reset(
+              owner_->buffer_pool_handle_.get(),
+              scratch.unique_page_ids[unique_index],
+              scratch.pages[unique_index] + offset_in_page);
+          continue;
+        }
+        *state.request->block =
```
richyreachy, line 733: 返回的 MemoryBlock 引用共享的线程局部缓冲区，下一次批量读取会覆盖或重新分配该缓冲区？

src/core/utility/buffer_storage.cc
iaojnh: 回退读取现在优先固定并读取驻留页；对于未驻留页，先同步加载/淘汰状态，再读取磁盘，避免丢失尚未刷盘的数据。

src/core/algorithm/hnsw/hnsw_algorithm.cc
iaojnh: 已修复正向和反向剪枝：存在外部 provider 时，通过距离计算器从 provider 获取向量，确保向量表示与当前 metric 一致；没有 provider 时保留原来的批量读取优化。

tools/core/local_builder.cc
iaojnh: 是的，已修复

src/core/utility/buffer_storage.cc
iaojnh: 已修复

richyreachy (approved):

PR #752, opened 2026-09-14
fix: correct inverted-index LIKE suffix and empty-array queries

zhourrr:
Fix two cases where inverted-index queries omit valid documents:

- Strip the leading wildcard before suffix lookup so patterns such as `LIKE '%.ts'` match correctly with extended wildcard enabled.
- Include empty arrays in non-null and negated-containment results for unsealed indexes by using the existing array-length index.

Regression tests check result counts and document IDs, including escaped suffix patterns and empty arrays across reopen and seal operations.

Validation: all 34 test cases across five relevant test targets passed locally.


egolearner (approved):
LGTM

PR #756, opened 2026-09-15
feat: relax name validation and harden recovery

### LINKED ISSUE #395: [Enhance]: Relax collection, field, and document ID name validation
### Affected Component

DB

### Current Behavior

The current name validation rules are very restrictive:

```c++
const std::regex COLLECTION_NAME_REGEX("^[a-zA-Z0-9_-]{3,64}$");

const std::regex FIELD_NAME_REGEX("^[a-zA-Z0-9_-]{1,32}$");

const std::regex DOC_PK_REGEX("^[a-zA-Z0-9_!@#$%+=.-]{1,64}$");
```

I don't see enough reasons to have such restrictions.

Image

### Desired Improvement

Lifting or relaxing these restrictions, or making them configurable, for:

- collection names
- field names
- document primary keys / IDs

A more permissive validation rule, or a user-configurable validator, would make Zvec easier to adopt in heterogeneous environments.

A more clear and user-friendly message indicating what is not allowed. 

### Impact

- Better compatibility with existing datasets and external identifiers
- Easier migration from other storage systems
- Reduced need for application-side name rewriting or ID normalization
- Less friction when integrating Zvec with multi-language tooling

zhourrr:
Relax document ID and collection name rules with centralized validation and clearer errors.

Fix related WAL recovery, schema, decoding, and Python/C binding issues while preserving storage formats.

Fixes #395 

## Validation Rules

The new rules are written as below, please review this first.

| Identifier | Previous rules | New rules |
|---|---|---|
| Document ID | 1–64 bytes; ASCII letters, digits, and `_!@#$%+=.-` | 1–1024 bytes; valid UTF-8, excluding C0/C1 control characters, DEL, and Unicode line/paragraph separators (U+2028/U+2029) |
| Collection name | 3–64 bytes; ASCII letters, digits, `_`, and `-` | 1–256 bytes; same UTF-8 and character restrictions as document IDs |
| Field name | 1–32 bytes; ASCII letters, digits, `_`, and `-` | 1–64 bytes; same ASCII character set, excluding five reserved names: `_zvec_row_id_`, `_zvec_g_doc_id_`, `_zvec_uid_`, `_zvec_score`, and `_zvec_group_id` |

### Summary

For valid ASCII inputs accepted by both implementations, the new validators were approximately **30–63× faster** than the previous `std::regex` checks in this microbenchmark.

### Valid Inputs Accepted by Both Implementations

| Input | Previous (ns/call) | New (ns/call) | Speedup |
|---|---:|---:|---:|
| Document ID, 8 ASCII bytes | 674.91 | 14.70 | 45.91× |
| Document ID, 32 ASCII bytes | 2,041.66 | 59.98 | 34.04× |
| Document ID, 64 ASCII bytes | 3,563.33 | 110.87 | 32.14× |
| Collection name, 8 ASCII bytes | 551.01 | 14.50 | 38.00× |
| Collection name, 32 ASCII bytes | 1,796.46 | 60.26 | 29.81× |
| Collection name, 64 ASCII bytes | 3,335.49 | 111.10 | 30.02× |
| Field name, 8 ASCII bytes | 653.86 | 11.68 | 55.98× |
| Field name, 16 ASCII bytes | 1,101.34 | 17.40 | 63.30× |
| Field name, 32 ASCII bytes | 1,824.54 | 28.75 | 63.46× |

### Newly Supported Inputs

These inputs were rejected by the previous implementation. Only the new validation cost is reported, since the two implementations perform different work.

| Input | New (ns/call) |
|---|---:|
| Document ID, 256 ASCII bytes | 420.46 |
| Document ID, 1,024 ASCII bytes | 1,793.87 |
| Collection name, 256 ASCII bytes | 436.82 |
| Field name, 64 ASCII bytes | 50.99 |
| Document ID, 28-byte mixed Unicode input | 42.13 |
| Collection name, 28-byte mixed Unicode input | 41.04 |


src/db/sqlengine/common/util.h
```diff
@@ -19,12 +19,10 @@
 
 namespace zvec::sqlengine {
 
-static const constexpr char *kFieldScore = "_zvec_score";
 static const constexpr char *kFieldVector = "_zvec_vector";
```
JalinWang: No refs. Remove them, as well as in the test?

zhourrr: done

JalinWang (dismissed):

JalinWang:
Should probably sync these to doc too

JalinWang (approved):

PR #759, opened 2026-09-15
fix(fts): persist sealed postings before dropping side column families

egolearner:
## Summary

If the process exits after a segment is dumped but before optimization finishes, recovery can reopen Roaring postings and the next optimize fails with `source postings is not BitPacked`. Converted postings bypass the RocksDB WAL and were not explicitly flushed before auxiliary column families were removed.

Synchronously flush converted postings and statistics before dropping auxiliary column families. Keep auxiliary data intact until removal, drop `$TF` first, and stop on any drop failure. Retries read actual column-family handles so cleanup can resume both in-process and after reopening. When `$TF` is missing, validate the postings format before continuing cleanup, rejecting legacy Roaring remnants instead of silently accepting them. This uses the existing column-family lifecycle without adding a persistent marker.

Related issue: https://github.com/zvec-ai/zvec-grep/issues/158

## Validation

- C++ `fts_column_indexer_test` and `fts_rocksdb_reducer_test` suites on macOS ARM64.
- Fault injection covers statistics write/flush failures and each auxiliary column-family drop, including retries after reopening.
- Process-exit tests cover both sealing paths and interruption before the first drop or after partial removal.
- Regression coverage rejects legacy Roaring postings with missing `$TF`.

Existing damaged indexes are not repaired by this change.


src/db/index/column/fts_column/fts_indexer.cc
```diff
@@ -308,21 +315,28 @@ Status FtsIndexer::seal_all() {
 
   // Convert all postings to bitpacked format.
   for (const auto &[name, indexer] : indexers_) {
+    if (converted_fields_.count(name) != 0) {
```
zhourrr: `converted_fields_` 仅保存在内存中，部分列族删除失败后的重试无法跨 `reopen` 恢复。
我用故障注入复现了以下流程：$TF 删除成功 → $MAX_TF 删除失败 → `seal_all()` 返回错误 → close/reopen → 再次 `seal_all()`。重新打开后，`converted_fields_` 为空，因此会再次执行转换，但 $TF 已不存在，最终报错：
```
FtsIndexer::seal_all convert failed: text
FtsColumnIndexer::convert_postings_to_bitpacked: not opened. field=text
```
由于第一次 seal 失败时，调用方尚未完成 writing segment 的切换，这会阻断重启后的 dump/optimize。建议持久化或在 open 时可靠恢复转换完成状态.

egolearner: 确实有这个问题，之前的内存状态无法覆盖关闭后重新打开的重试场景。

已调整为：保留辅助数据直到删除列族，先同步刷盘 postings 和统计信息，再优先删除 `$TF`，最后清理剩余列族。这样无需新增持久化标记，也能在部分删除失败或重启后继续清理。`$TF` 缺失时还会检查 postings 格式，避免误判旧版本的损坏数据。

zhourrr (approved):

PR #761, opened 2026-09-16
feat(search): add fast_query for dense read-only collections

luoxiaojian:
Return owned internal-ID arrays and optional scores through C++ search_fast and Collection.fast_query with a thin NumPy binding. Reuse query parameters and graph/Flat fast paths, retaining filtered and multi-segment fallbacks.

Preserve the existing scale_factor defaults and Uniform raw-Flat fallback. Cover native reference types, default and explicit refine budgets, all dense index families, parameter reuse, result ownership, and untrained/mixed Uniform segments.

Validation: 634 related Python tests passed; 70 C++ tests passed with one existing skip. Ruff, changed-line clang-format, and diff checks passed.

Adapted from alignment commit 9ddbd5aa082615226d91addb41cd60496143d15a onto upstream/main 43f72b186a0524b7f6b53c645bf90ac44ac05db3.


tests/ailego/buffer/vector_page_table_test.cc
```diff
@@ -727,7 +727,9 @@ TEST_F(BufferPoolTest, ConcurrentWritablePressureUsesBackgroundWriteback) {
 }
 
 TEST_F(BufferPoolTest, RecoversDirtyPageAfterQueueRegistrationFailure) {
-  InitTablePool(/*capacity_pages=*/1, /*entry_num=*/1);
```
richyreachy: @iaojnh 这个UT 文件里还有一些case的InitTablePool，也需要做一下设置

src/db/index/segment/segment.cc
```diff
@@ -4534,6 +4540,30 @@ BlockID SegmentImpl::allocate_block_id() {
...
+  for (size_t i = 0; i < doc_ids_.size(); ++i) {
+    if (doc_ids_[i] != i) return false;
+  }
+  return true;
+}
+
+Status SegmentImpl::get_global_doc_ids(const std::vector<int> &segment_doc_ids,
+                                       std::vector<int64_t> &out) const {
+  out.resize(segment_doc_ids.size());
+  std::lock_guard lock(seg_mtx_);
```
JalinWang: Probably don't need an exclusive lock here?

src/db/collection.cc
```diff
@@ -1805,6 +1840,240 @@ Result<DocPtrList> CollectionImpl::query(const SearchQuery &query) const {
...
+  CHECK_CLOSED_RETURN_STATUS_EXPECTED(closed_, false);
+  if (!options_.read_only_) {
+    return tl::make_unexpected(
+        Status::InvalidArgument("fast query requires a read-only collection"));
+  }
+  auto cached = fast_query_states_.find(field_name);
+  if (cached == fast_query_states_.end()) {
+    auto resolved = resolve_fast_query(field_name);
+    if (!resolved) return tl::make_unexpected(resolved.error());
+    cached = fast_query_states_.emplace(field_name, std::move(resolved.value()))
```
JalinWang: Might need to lock the shared `fast_query_states_ ` here?

src/db/index/segment/segment.cc
luoxiaojian: 改了一下，先看看

PR #766, opened 2026-09-17
feat(buffer): integrate IVF and DiskANN with shared page cache

### LINKED ISSUE #704: [Feature] Launch Buffer Storage for Memory-Bounded Query Caching

iaojnh:
## Summary

- Add Buffer Storage support for IVF and DiskANN.
- Reduce IVF posting-list copying through page-span reads.
- Cache DiskANN graph pages with batched direct-I/O fallback under cache pressure.
- Handle bypass-only budgets, preserve backing-file identity, and fix cache/I/O resource lifetimes.

Existing mmap and direct-I/O modes remain available.

## Testing

- macOS: 298 tests and 920 targeted regression executions passed.
- clang-format and clang-tidy passed.
- Linux and Windows validation pending CI.


src/core/algorithm/ivf/ivf_index_provider.h
```diff
@@ -89,17 +89,19 @@ class IVFIndexProvider : public IndexProvider {
...
     const void *data() const override {
-      return entity_->get_vector(current_local_id());
+      size_t local_id = current_local_id();
+      return local_id < count_ ? entity_->get_vector(local_id) : nullptr;
     }
 
     //! Test if the iterator is valid
     bool is_valid() const override {
-      return pos_ < count_;
+      return pos_ < count_ && (!use_mapping_ || ensure_mapping_chunk());
```
richyreachy: ensure_mapping_chunk() 读取失败，不应该和其它错误混在一起，误判为invalid？

iaojnh: 已修复

richyreachy (approved):

PR #690, opened 2026-08-20
refactor(sqlengine): add extensible filter set rewriting

egolearner:
## Summary

- Centralize recursive filter normalization and introduce registries for group and set rewrite rules.
- Fix the incorrect rewrite of OR-connected not-equal predicates into NOT IN.
- Merge same-field equality/IN predicates under OR and inequality/NOT IN predicates under AND.
- Deduplicate exact literals in merged results using string_view keys while leaving standalone IN/NOT IN lists unchanged.
- Keep positive intersections under AND and semantic value canonicalization outside the initial rule set.
- Add regression tests for rewrite boundaries, invalid literals, duplicate handling, predicate ordering, and unsatisfiable-filter propagation.

## Testing

- ninja -C build.release unittest.zvec_sqlengine
- clang-format -n --Werror on changed C++ files
- git diff --check

PR #769, opened 2026-09-18
refactor(sqlengine): separate filter validation and execution binding

egolearner:
Follow-up to https://github.com/alibaba/zvec/pull/690.

Separate search condition validation, rewriting, and execution binding so rewrite rules can simplify the AST without carrying stale execution state. Preserve the previous input acceptance rules, including existing numeric conversion and function validation behavior.

- Split the original walker into dedicated Validator and Binder files. Validate original predicates before rewriting and resolve fields directly from the schema without a cross-stage metadata cache.
- Determine final forward, inverted, and vector execution subtrees before converting constants. Keep forward literals in their original form and remove empty logical wrappers after extracting conditions.
- Resolve function result types explicitly, preserve detailed binding errors, and repair literal parent pointers when rewrite rules merge predicates into lists.
- Cover rewrite boundaries, legacy validation compatibility, replacement nodes, function binding, and mixed execution paths with regression tests.

Validation: nine related unit-test targets passed (209 tests): simple_rewriter_test, query_info_test, contain_test, like_test, forward_recall_test, invert_recall_test, vector_recall_test, sqlengine_test, and optimizer_test. The 70 simple_rewriter_test cases passed again after restoring the original get_set_field parameter. Formatting and git diff --check passed.


src/db/sqlengine/analyzer/search_cond_binder.cc
```diff
@@ -0,0 +1,310 @@
...
+  const auto &root = query_info->search_cond();
+  if (root == nullptr) {
+    return Status::OK();
+  }
+  auto status = classify(root, false);
+  if (!status.ok()) {
+    return status;
+  }
+  if (filter_count_ > kMaxNumOfFilters) {
+    return Status::NotSupported("max number of filters is limited to 4096");
```
zhourrr: 可以考虑改下 error message："too many filter conditions: <n>; the maximum is <kMaxNumOfFilters>"
"vector search condition cannot be combined with OR"

这样用户会得到 ：“ValueError: Analyze SQL info failed: too many filter conditions: 5000; the maximum is 4096”

egolearner: 已调整。
下面的改为了"vector search condition cannot appear within an OR expression"，防止用户误以为向量条件不能和带or的filter一起使用。

zhourrr (approved):

